PR #12, opened 2025-12-04
Privi listens to notifications from jetstream

eagraf:
Privi listens to notifications from jetstream

Reorganize a bit

Fix gitignore rb

go mod tidy

Testing

PR #11, opened 2025-12-04
dx tweaks

eagraf:
Mostly small fixes for things that I ran into when re-setting up the repo for the first time.

PR #13, opened 2025-12-04
Ingest notification events

eagraf:
When processing new notification events, insert them into the database

Do some refactoring

Comment

Tests

### DIFF
--- a/internal/privi/inbox_test.go
+++ b/internal/privi/inbox_test.go
@@ -17,10 +17,8 @@ func TestNotificationIngester(t *testing.T) {
 	db, err := gorm.Open(sqlite.Open(":memory:"), &gorm.Config{})
 	require.NoError(t, err)
 
-	err = db.AutoMigrate(&Notification{})
+	ingester, err := NewNotificationIngester(db)
 	require.NoError(t, err)
-
-	ingester := NewNotificationIngester(db)
 
 	t.Run("ignores non-commit events", func(t *testing.T) {
 		event := &models.Event{


eagraf:

- #11
    - #12
        - #13 ◀
            - #14
                - #16
                    - #17
                        - #18
                            - #19
                                - #20
                                    - #21
            - #15

cmd/privi/main.go
```diff
@@ -168,6 +171,19 @@ func setupDB(dbPath string) *gorm.DB {
 		log.Fatal().Err(err).Msg("unable to open sqlite file backing privi server")
 	}
 
+	err = priviDB.AutoMigrate(&privi.Notification{})
```
sashankg: these should go in `NewNotificationIngester`

internal/privi/inbox.go
```diff
@@ -3,13 +3,52 @@ package privi
...
+
+// Notification is a Gorm model for notifications.
+type Notification struct {
+	gorm.Model
+	Did        string
+	OriginDid  string
+	Collection string
+	Rkey       string
+	CreatedAt  time.Time `gorm:"autoCreateTime"`
+	UpdatedAt  time.Time `gorm:"autoUpdateTime"`
```
sashankg: gorm.Model has these

cmd/privi/main.go
```diff
@@ -75,6 +75,8 @@ func run(_ context.Context, cmd *cli.Command) error {
 
 	// Start the notification listener in a separate goroutine
 	eg.Go(func() error {
+		// Note that for now, we're ingesting all notifications in the entire system
```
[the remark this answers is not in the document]
eagraf: That only filters for the notifications, but we will still be processing every notification from every PDS even when not relevant

cmd/privi/main.go
sashankg: idk to me it makes sense to call AutoMigrate in the `New*` function for a component. that's where you setup all the requirements for the component so to me it's not additional cognitive load. From then on you're guaranteed that the model is available to use within the component. If you move the AutoMigrate call out of `New*` then the component just needs to trust that it was called earlier which seems more brittle.

When I said "calling AutoMigrate multiple times" I just meant that if two components need the same model type then they can each call AutoMigrate on the model so they don't need to depend on someone else to do it.

I don't have a case against your point about "one of the first things I'd like to see off the bat is what gets stored all in one place". that's true but to me it's more important to see which components use which models and declaring it in the constructor achieves that.

I personally don't like a central models file because there isn't any semantic relationship between them besides all being models.

If you're still unconvinced, I'm fine to disagree and commit to unblock as long as we're on the same page about not in privi/main.go

eagraf: Ok going to go with your way, because even if we centralized the migration commands, we would also need to move the actual model structs for the centralized approach to make sense

PR #14, opened 2025-12-04
Oauth small fixes

eagraf:


eagraf:

- #11
    - #12
        - #13
            - #14 ◀
                - #16
                    - #17
                        - #18
                            - #19
                                - #20
                                    - #21

internal/oauthserver/oauth_server.go
```diff
@@ -33,6 +33,7 @@ type authRequestFlash struct {
 	Form           url.Values // Original authorization request form data
 	DpopKey        []byte
 	AuthorizeState *oauthclient.AuthorizeState // AT Protocol authorization state
+	Did            syntax.DID                  // DID of the user
```
sashankg: this is probably fine but we should try to keep the flash small so ideally we don't store both the handle from url.Values and the did

sashankg (approved):

PR #27, opened 2025-12-14
[Typescript] Pull out AuthForm

sashankg:


typescript/internal/src/AuthForm.tsx
```diff
@@ -0,0 +1,49 @@
...
+    isPending,
+    error,
+    isError,
+  } = useMutation({
+    async mutationFn({ handle }: AuthFormData) {
+      if (!handle) {
+        throw new Error("Handle required");
+      }
+      const url = authManager.loginUrl(handle, redirectUrl);
+      window.location.href = url.toString();
```
eagraf: Is there a reason to use this rather than the routing framework to change the page location?
 Might be a bit tough because that could vary based on what is importing this component.

sashankg: the routing framework is irrelevant here. We need to do an external page nav which doesn't depend on the routing framework at all.

PR #29, opened 2025-12-16
[Moon] Use template to generate web app

sashankg:
Can generate a new web app by doing 

```
moon generate web-app -- --name ...
```

nice because we can continue to update the template and regenerate. it will just replace the template files. We can move more shared logic into `internal/` over time.

### DIFF
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -6,24 +6,9 @@ settings:
 
 catalogs:
   default:
-    '@tanstack/devtools-vite':
-      specifier: ^0.3.11
-      version: 0.3.12
-    '@tanstack/react-devtools':
-      specifier: ^0.7.0
-      version: 0.7.11
     '@tanstack/react-query':
       specifier: ^5.90.12
-      version: 5.90.12
-    '@tanstack/react-router':
-      specifier: ^1.132.0
-      version: 1.141.2
-    '@tanstack/react-router-devtools':
-      specifier: ^1.132.0
-      version: 1.133.32
-    '@tanstack/router-plugin':
-      specifier: ^1.132.0
-      version: 1.141.2
+      version: 5.90.16
     '@testing-library/dom':
       specifier: ^10.4.0
       version: 10.4.1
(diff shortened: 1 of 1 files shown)


templates/web-app/moon.yml
```diff
@@ -0,0 +1,8 @@
+tasks:
+  dev:
+    env:
+      DOMAIN: "$project.${TS_DOMAIN}"
+    deps:
+      - target: "privi:dev"
+      - target: "funnel:start"
+        args: ["5173", "$project"]
```
eagraf: Lets use a port far away from 5173 (6000 might be good). 5173 is the same as our base frontend, and there are probably testing scenarios where we want an app frontend and the Habitat frontend running together. 5174 is not a good option because vite will increment the port number, so its easy for there to be conflicts all over the place

templates/web-app/src/main.tsx
```diff
@@ -0,0 +1,59 @@
...
+  createHashHistory,
+  createRouter,
+} from "@tanstack/react-router";
+// Import the generated route tree
+import { routeTree } from "./routeTree.gen";
+import { QueryClient, QueryClientProvider } from "@tanstack/react-query";
+import reportWebVitals from "internal/reportWebVitals.ts";
+import { AuthManager } from "internal/authManager.ts";
+
+const authManager = new AuthManager(__DOMAIN__, __HABITAT_DOMAIN__, () => {
```
eagraf: Note that having HABITAT_DOMAIN set should be optional. I think we still need it for now because you might want to test on a account hosted on bsky, but still use privi.  If its not set in the env, we should discover the pds url from the did doc of the logged in user.

templates/web-app/template.yml
```diff
@@ -0,0 +1,12 @@
+# https://moonrepo.dev/docs/config/template
+$schema: '../../.moon/cache/schemas/template.json'
+
+title: 'Web App'
```
eagraf: Lets give these some habitat specific info

eagraf (approved):

PR #31, opened 2025-12-18
[Docs] Multiplayer POC

sashankg:
Quick demo of multiplayer frontend only apps. All habitat needs to provide is relay services for pubsub topics. We can extend it to provide relay transport such that apps can start talking directly to each other and take some load off the backend. 

https://github.com/user-attachments/assets/d8748374-e6bc-4472-a4ac-fc3b4881c276

TODO/Open questions:
- Right now all peers are listening and writing to the same doc. Only works if all peers are authed as the owner of the doc.
  - I still think habitat shouldn't manage write permissions. That can be delegated to the app layer. For collaborating on a record, "writers" will have their own copy (it just needs to be their diff from the true record). They can use the notification system to tell the owner about updates. The owners app will consume the notification and update the owner record. 
- Auth and permissions might be complicated but there probably is a way out. If habitat is required to acts as the relay, we can do permission checks there and block unauthorized peers from subscribing to a topic. However I think the more robust solution is to give apps an ephemeral verifiable token to prove it has read permission for a record. Peers would use that to authorize connections between each other. 
- Topics should probably always be at atUris and the permissions to subscribe to a topic should inherit from the permissions of that record. Or if we support tokens that the app can use to verify permissions, we can leave it completely open to the app.

### DIFF
--- a/internal/p2p/server.go
+++ b/internal/p2p/server.go
@@ -17,8 +17,8 @@ import (
 )
 
 type Server struct {
-	host        host.Host
-	proxy       *httputil.ReverseProxy
+	host  host.Host
+	proxy *httputil.ReverseProxy
 }
 
 var _ io.Closer = (*Server)(nil)
@@ -57,8 +57,8 @@ func NewServer() (*Server, error) {
 	}
 	url.Scheme = "http"
 	return &Server{
-		host:        host,
-		proxy:       httputil.NewSingleHostReverseProxy(url),
+		host:  host,
+		proxy: httputil.NewSingleHostReverseProxy(url),
 	}, nil
 }
 


cmd/privi/main.go
```diff
@@ -66,6 +67,10 @@ func run(_ context.Context, cmd *cli.Command) error {
 	oauthServer := setupOAuthServer(keyFile, domain)
 	priviServer := setupPriviServer(db, oauthServer)
 	pdsForwarding := newPDSForwarding(oauthServer)
+	p2pServer, err := p2p.NewServer()
```
arushibandi: oh interesting i didn't realize we would host a `p2p` server

typescript/apps/docs/src/connectionProvider.ts
```diff
@@ -0,0 +1,136 @@
...
+  ) => any;
+  status: (event: {
+    status: "connected" | "disconnected" | "connecting";
+  }) => any;
+  "connection-error": (event: Event, provider: Libp2pConnectionProvider) => any;
+  sync: (state: boolean) => any;
+}> {
+  node: Node;
+  doc: Y.Doc;
+  awareness: awarenessProtocol.Awareness;
```
arushibandi: i didn't even know about this

typescript/apps/docs/src/connectionProvider.ts
```diff
@@ -0,0 +1,136 @@
...
+    provider: Libp2pConnectionProvider,
+  ) => any;
+  status: (event: {
+    status: "connected" | "disconnected" | "connecting";
+  }) => any;
+  "connection-error": (event: Event, provider: Libp2pConnectionProvider) => any;
+  sync: (state: boolean) => any;
+}> {
+  node: Node;
+  doc: Y.Doc;
```
arushibandi: should the doc not live outsie the "Libp2pConnectionProvider"? like a higher level? and it uses the connection provider?

PR #34, opened 2025-12-24
Add DID to sub field in token oauth response

eagraf:

PR #36, opened 2025-12-24
Fix hot reloading for privi and logs

eagraf:


eagraf:

- #33
    - #34
        - #35
            - #36 ◀
                - #37
                    - #38

.moon/tasks/go.yml
```diff
@@ -48,3 +50,4 @@ tasks:
     options:
       envFile:
         - "/dev.env"
+      outputStyle: "stream"
```
sashankg: this is automatic when the task name is `dev` and when `local: true`: https://moonrepo.dev/docs/config/project#local

.moon/tasks/go.yml
```diff
@@ -37,6 +37,8 @@ tasks:
 
   start:
     command: "$workspaceRoot/bin/$project"
+    deps:
```
sashankg: I split this up on purpose because I wanted them to be separate steps in the dockerfile for the integration test. Agree it's not ideal but shouldn't really get in the way for dev, right?

cmd/funnel/moon.yml
```diff
@@ -0,0 +1,5 @@
+tasks:
+  start:
+    options:
+      outputStyle: "none"
```
sashankg: why? I think if anything we should control the amount of logs in the script itself via an env var

cmd/privi/.air.toml
sashankg: imo we should just have one `.air.toml` file at the root which handles include and exclude dir and reuse it across go binaries

cmd/privi/moon.yml
```diff
@@ -12,6 +12,8 @@ tasks:
       - target: "funnel:start"
         args: ["5173", "frontend"]
       - target: "frontend:start"
+    options:
```
sashankg: same don't need this

cmd/privi/.air.toml
sashankg: i think i did something here: https://github.com/habitat-network/habitat/pull/31/files. If we run in workspace root it'll update on any go file change. probably add a config file in the root with all the right ignores

PR #51, opened 2026-01-17
Add notifications endpoints to client

eagraf:
Add notifications endpoints to client

Use proper types in client

Some fixes


sashankg (approved):
approving since no actual problem with the diff. However, now that we have generated types, we probably don't need a client where we redefine everything. We can just make a generic helper function that takes the input type and output type as type params.

PR #54, opened 2026-01-17
Make sure sub field is set in session

eagraf:


sashankg (changes requested):
oh wait is this just to set the sub in `HandleToken` response extras? if so then we should just stop doing that. the client can parse the jwt to get the subject

PR #55, opened 2026-01-17
[observability]: setup opentelemetry <> grafana

arushibandi:
Out of the box set up for OpenTelemetry <> Grafana by following the instructions here: https://grafana.com/docs/opentelemetry/instrument/go/

Added a single metric that records when a process starts running for proof-of-concept (to see it show up in the Grafana dashboard). [See it here](https://habitatnetwork.grafana.net/d/arktc98/arushi-s-first-dashboard?from=now-15m&to=now&timezone=browser). (Should be viewable once accepting the invite to join my Grafana team).

I also added the relevant env vars / secrets into Google cloud. So we should start seeing metrics from the GCP instance once this goes out.

Follow-ups:
- Get logs to show up in Grafana
- Add actual metrics
- Figure out tagging for things like `version`, `service` , `env` etc.
- Setup Go profiling with Grafana


cmd/privi/otel.go
sashankg: if none of this privi specific, we should move it to internal

PR #67, opened 2026-01-24
[OAuthClient] Refresh PDS credentials on expiry

sashankg:
will work on an integration test with a local pds but should be able to merge this now

### DIFF
--- a/internal/oauthclient/dpop_test.go
+++ b/internal/oauthclient/dpop_test.go
@@ -259,19 +259,12 @@ func TestAuthedDpopHttpClient(t *testing.T) {
 	}))
 	defer server.Close()
 
-	// Create test key
-	key, err := ecdsa.GenerateKey(elliptic.P256(), rand.Reader)
-	require.NoError(t, err)
+	clientFactory := NewPDSClientFactory(testPdsCredStore(t, jwt.Claims{
+		Expiry: jwt.NewNumericDate(time.Now().Add(10 * time.Minute)),
+	}), testOAuthClient(t), NewDummyDirectory(server.URL))
 
-	// Create DPoP client
-	client := newAuthedDpopHttpClient(
-		testIdentity(server.URL),
-		testPdsCredStore(t, key, jwt.Claims{
-			Expiry: jwt.NewNumericDate(time.Now().Add(10 * time.Minute)),
-		}),
-		testOAuthClient(t),
-		&MemoryNonceProvider{nonce: "test-nonce"},
-	)
+	client, err := clientFactory.NewClient(t.Context(), testIdentity(server.URL).DID)
+	require.NoError(t, err)
 
 	// Create request with Authorization header
 	req, err := http.NewRequest("GET", "/test", nil)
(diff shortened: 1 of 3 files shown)


internal/oauthclient/clientFactory.go
```diff
@@ -0,0 +1,39 @@
...
+
+func (f *PDSClientFactory) NewClient(
+	ctx context.Context,
+	did syntax.DID,
+) (*AuthedDpopHttpClient, error) {
+	id, err := f.dir.LookupDID(ctx, did)
+	if err != nil {
+		return nil, fmt.Errorf("failed to lookup did: %w", err)
+	}
+	return NewAuthedDpopHttpClient(id, f.credStore, f.oauthClient, &MemoryNonceProvider{}), nil
```
arushibandi: this is a bad pattern where we pass in a whole credStore + oauthClient to a struct that is created for a single user and is also ephemeral. It makes me feel like instead of `AuthedDpopHttpClient` which only has one `Do` method, we should just bake that all into a single method off of the client factory, like `s.pdsClientFactory.MakeRequest(r.Context(), did, req)` instead of
```
dpopClient, err := s.pdsClientFactory.NewClient(r.Context(), did)
	if err != nil {
		utils.LogAndHTTPError(w, err, "create dpop client", http.StatusInternalServerError)
		return
	}
	req, err := http.NewRequest(
		http.MethodPost,
		"/xrpc/com.atproto.repo.createRecord",
		bytes.NewReader(body),
	)
	if err != nil {
		utils.LogAndHTTPError(w, err, "create request", http.StatusInternalServerError)
		return
	}
	req.Header.Set("Content-Type", "application/json")

	resp, err := dpopClient.Do(req)
	if err != nil {
		utils.LogAndHTTPError(w, err, "forward request", http.StatusInternalServerError)
		return
	}
	defer func() { _ = resp.Body.Close() }()
```

internal/oauthclient/clientFactory.go
```diff
@@ -0,0 +1,39 @@
...
+		oauthClient: oauthClient,
+		dir:         dir,
+	}
+}
+
+func (f *PDSClientFactory) NewClient(
+	ctx context.Context,
+	did syntax.DID,
+) (*AuthedDpopHttpClient, error) {
+	id, err := f.dir.LookupDID(ctx, did)
```
arushibandi: it might also be nice to take in an `*identity.Identity` in more places rather than a `did`.

sashankg: that requires every callsite doing the lookup and needing to depend on directory. I thought it was nicer to only do the lookup where needed and just have the factory depend on directory

internal/oauthclient/clientFactory.go
sashankg: that's the primary reason for organizing this way

arushibandi: I don't see how the code above couldn't be factored into this:

```
req, err := http.NewRequest(
		http.MethodPost,
		"/xrpc/com.atproto.repo.createRecord",
		bytes.NewReader(body),
	)
	if err != nil {
		utils.LogAndHTTPError(w, err, "create request", http.StatusInternalServerError)
		return
	}
	req.Header.Set("Content-Type", "application/json")

resp, err := s.pdsClientFactory.MakeAuthedRequest(r.Context(), did, req)
	if err != nil {
		utils.LogAndHTTPError(w, err, "forward request", http.StatusInternalServerError)
		return
	}
	defer func() { _ = resp.Body.Close() }()
```

PR #72, opened 2026-02-01
getRecord returns 501 not implemented if it needs to forward to another Habitat node

eagraf:
getRecord returns 501 not implemented if it needs to forward to another Habitat node

Refactor so that the forwarding branch is inside the store layer

### DIFF
--- a/internal/pear/pear_test.go
+++ b/internal/pear/pear_test.go
@@ -100,12 +100,12 @@ func TestGetRecordForwardingNotImplemented(t *testing.T) {
 	require.NoError(t, err)
 	repo, err := NewSQLiteRepo(db)
 	require.NoError(t, err)
-	p := newStore(perms, repo)
+	p := newPermissionEnforcingRepo(perms, repo)
 
 	// Try to get a record for a DID that doesn't exist on this server
 	got, err := p.getRecord("some.collection", "some-rkey", "did:plc:unknown123", "did:plc:caller456")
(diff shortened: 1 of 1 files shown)


eagraf:

- #72 ◀
    - #73

internal/privi/server.go
```diff
@@ -181,11 +181,29 @@ func (s *Server) GetRecord(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 
+	has, err := s.store.hasRepoForDid(targetDID.String())
```
sashankg: why do it twice? just let `store.getRecord` fail, right?

internal/privi/privi.go
```diff
@@ -61,6 +62,15 @@ func (p *store) getRecord(
 	targetDID syntax.DID,
 	callerDID syntax.DID,
 ) (*Record, error) {
+
+	has, err := p.hasRepoForDid(targetDID.String())
+	if err != nil {
+		return nil, err
+	}
+	if !has {
+		return nil, ErrForwardingNotImplemented
```
arushibandi: I don't think we should return  `ErrForwardingNotImplemented`, we should return `ErrRepoNotFound`.

We shouldn't forward -- our setup will be one of two things:
either clients will use the at-proto proxy header to find the right habitat server to talk to OR we just have a global `habitat.network` API that everyone calls and it does routing in the back.

PR #74, opened 2026-02-02
Add collection column to record table

eagraf:
Adds a collection column to record table, for easier querying in the future

Note this doesn't break down the concatenation in the rkey column (its currently collection.rkey), but we should do this in the future.

### DIFF
--- a/internal/pear/pear.go
+++ b/internal/pear/pear.go
@@ -44,7 +44,7 @@ func (p *permissionEnforcingRepo) putRecord(
 	validate *bool,
 ) error {
 	// It is assumed right now that if this endpoint is called, the caller wants to put a private record into pear.
-	return p.repo.putRecord(did, fmt.Sprintf("%s.%s", collection, rkey), record, validate)
+	return p.repo.putRecord(did, collection, rkey, record, validate)
 }
 
 // getRecord checks permissions on callerDID and then passes through to `repo.getRecord`.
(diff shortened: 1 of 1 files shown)


sashankg:
> Note this doesn't break down the concatenation in the rkey column (its currently collection.rkey), but we should do this in the future.

let's just do this now as part of this change

internal/privi/repo.go
```diff
@@ -33,9 +33,10 @@ type sqliteRepo struct {
 }
 
 type Record struct {
-	Did  string `gorm:"primaryKey"`
-	Rkey string `gorm:"primaryKey"`
-	Rec  string
+	Did        string `gorm:"primaryKey"`
+	Collection string `gorm:"primaryKey"`
+	Rkey       string `gorm:"primaryKey"`
+	Rec        string
```
arushibandi: should we be storing this as bytes?

also we should cal this `Value`

internal/privi/repo.go
```diff
@@ -171,37 +181,96 @@ func (r *sqliteRepo) listRecords(
...
+				specificRkeys = append(specificRkeys, a)
 			}
 		}
-		query = query.Where(allowConditions)
+
+		// If we have a wildcard, no need to filter by rkey (collection filter is enough)
+		// Otherwise, filter by specific rkeys
+		if !hasWildcard && len(specificRkeys) > 0 {
+			query = query.Where("rkey IN ?", specificRkeys)
+		}
```
arushibandi: this is an optimization right? so we don't need to post-filter the results?

PR #75, opened 2026-02-02
Try to fetch notified record immediately upon getting a notification

eagraf:
Try to fetch notified record immediately upon getting a notification

Fix some tests

### DIFF
--- a/internal/privi/inbox.go
+++ b/internal/privi/inbox.go
@@ -22,10 +22,10 @@ type notificationRecord struct {
 // Notification is a Gorm model for notifications.
 type Notification struct {
 	gorm.Model
-	Did                 string
-	OriginDid           string
-	Collection          string
-	Rkey                string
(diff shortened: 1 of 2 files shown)


internal/privi/inbox.go
```diff
@@ -100,16 +108,64 @@ func (n *NotificationIngester) createNotification(did string, originDid string,
...
+	}
+	// Query the origin repo with us as the caller
+	// If the record fetch fails for any reason (forwarding not implemented, record not found,
+	// unauthorized, etc.), we set the value to an empty string and still create the notification.
+	// This ensures notifications are created even when we can't fetch the shadowed record.
+	recordMarshalledJson := ""
+	fetchFailed := false
+	var lastSuccessfulFetch *time.Time
+
+	record, err := n.store.GetRecord(collection, rkey, origin, caller)
```
sashankg: the "get record" needed here has different semantics than what the store should implement. We should have a separate component that handles the message channel between nodes which the store and the notification ingester would depend on

PR #78, opened 2026-02-03
Create a simple user store

eagraf:
This PR just adds a table which lists user dids managed by the node. We need this because the old way of checkign broke if you had no previous activity on this node.

The did is written into the table after your first login.


internal/pear/sqlite_repo.go
```diff
@@ -29,7 +31,8 @@ import (
 // We really shouldn't have unexported types that get passed around outside the package, like to `main.go`
 // Leaving this as-is for now.
 type repo struct {
-	db *gorm.DB
+	db        *gorm.DB
+	userStore userstore.UserStore
```
arushibandi: let's ship it for now but 

i'm wondering if this should just be a part of the repo ... basically just have the ensureUser / checkUserExists methods on the repo and have  `permissionsEnforcingRepo` implement the `UserStore` interface you have

PR #81, opened 2026-02-07
WIP cross repo record listing

eagraf:
listRecords queries across all repos on the node that for a collection. If you have permission for any of the other user's repos records, they will be merged into the result.

### DIFF
--- a/internal/pear/pear.go
+++ b/internal/pear/pear.go
@@ -92,12 +92,12 @@ func (p *Pear) getRecord(
 		return nil, err
 	}
 
-	if authz {
-		// User has permission, return the record
-		return p.repo.getRecord(targetDID.String(), collection, rkey)
+	if !authz {
+		return nil, ErrUnauthorized
 	}
 
-	return nil, ErrUnauthorized
+	// User has permission, return the record
+	return p.repo.getRecord(targetDID.String(), collection, rkey)
 }
 
 func (p *Pear) listRecords(
@@ -107,7 +107,7 @@ func (p *Pear) listRecords(
 ) ([]Record, error) {
 	var allRecords []Record
 
-	// Step 2: Get records from caller's own repo
+	// Step 1: Get records from caller's own repo
 	allow, deny, err := p.permissions.ListReadPermissionsByUser(
 		did.String(),
 		callerDID.String(),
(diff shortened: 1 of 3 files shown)


internal/pear/pear.go
```diff
@@ -48,61 +49,208 @@ func (p *permissionEnforcingRepo) putRecord(
 }
 
 // getRecord checks permissions on callerDID and then passes through to `repo.getRecord`.
+// It also searches through notifications for records the user has access to, even if they don't belong to the targetDID.
```
sashankg: why do we need check notifications? can't we just forward the request to the owner's node?

arushibandi (commented):
the overall PR looks good, but i think you need to replace notifications with the inbox now that we changed some stuff

internal/pear/pear.go
eagraf: Might be unnecessary, but I think its reasonable to check that you have permissions for something on your end first

sashankg: it's just extra work, right? either way the record owner will validate the requestor's permission. also, the inbox is designed to be eventually consistent so may not have a notification for a record that the user does have access to.

internal/pear/pear.go
```diff
@@ -129,7 +117,54 @@ func (p *Pear) listRecords(
...
+	ownRecords, err := p.repo.listRecords(callerDID.String(), collection, allow, deny)
+	if err != nil {
+		return nil, err
+	}
+	allRecords = append(allRecords, ownRecords...)
+
+	// Step 3: Query permissions store to get two lists:
+	// 1. Users with full collection access
+	// 2. Specific records with direct permissions
+	fullAccessOwners, err := p.permissions.ListFullAccessOwnersForGranteeCollection(callerDID.String(), collection)
```
eagraf: IMO everything here is very janky, but doing this cleanly with the way the permissionst table is set up right now is pretty difficult

sashankg (commented):
we need to cut down on the number of queries we're making. we should just need to make 2: 1 to the permission store and 1 to the repo. The permission store query should return a full list where each item is owner + collection + rkey/*. the query to the repo should be constructed using the permissions list as filters.

sashankg (approved):

PR #83, opened 2026-02-07
Add grantees  to habitat client

eagraf:


arushibandi (approved):
we could possible add stricter typing and type this as DIDGrantee | CliqueGrantee with some regex validation. but lg for now

PR #86, opened 2026-02-08
[pear]: implement notify of update

arushibandi:
New implementation of inbox (does not use jetstream).
Fix up notifyOfUpdate lexicon.
Simple implementation of inbox -- basically one db table hosting what used to be notifications.

### DIFF
--- a/cmd/pear/main.go
+++ b/cmd/pear/main.go
@@ -246,7 +246,11 @@ func setupPriviServer(
 		log.Fatal().Err(err).Msg("unable to setup permissions store")
 	}
 
-	return pear.NewServer(inbox.NewInbox(db), permissionStore, repo, oauthServer, pdsClientFactory)
+	inbox, err := inbox.NewInbox(db)
+	if err != nil {
+		log.Fatal().Err(err).Msg("unable to setup inbox")
+	}
+	return pear.NewServer(inbox, permissionStore, repo, oauthServer, pdsClientFactory)
 }
 
 func setupOAuthServer(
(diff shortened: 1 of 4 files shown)


internal/pear/pear.go
```diff
@@ -26,10 +32,11 @@ var (
 	ErrUnauthorized            = fmt.Errorf("unauthorized request")
 )
 
-func newPermissionEnforcingRepo(perms permissions.Store, repo *repo) *permissionEnforcingRepo {
+func newPermissionEnforcingRepo(db *gorm.DB, perms permissions.Store, repo *repo) *permissionEnforcingRepo {
```
sashankg: i want to refactor all this to make it simpler but I think the technically correct dependency injection thing to do here is to take an Inbox as input not the db.

lexicons/network/habitat/internal/notifyOfUpdate.json
sashankg: let's merge https://github.com/habitat-network/habitat/pull/87 and then regenerate

lexicons/network/habitat/internal/notifyOfUpdate.json
```diff
@@ -11,18 +11,29 @@
                     "type": "object",
                     "required": [
                         "collection",
-                        "did"
+                        "rkey",
+                        "sender",
+                        "recipient"
                     ],
                     "properties": {
+                        "sender": {
```
sashankg: do we need sender, here? it's redundant with the auth token, right

sashankg (approved):

PR #90, opened 2026-02-08
[XrpcChannel] Init

sashankg:
Created an abstraction around inter-node communication. For now it just sends a request to the recipient pear node via the sender's PDS.

### DIFF
--- a/cmd/pear/pds_forwarding.go
+++ b/cmd/pear/pds_forwarding.go
@@ -12,7 +12,7 @@ import (
 
 type pdsForwarding struct {
 	oauthServer      *oauthserver.OAuthServer
-	pdsClientFactory *oauthclient.PDSClientFactory
+	pdsClientFactory oauthclient.PDSClientFactory
 }
 
 var _ http.Handler = (*pdsForwarding)(nil)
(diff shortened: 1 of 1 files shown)


internal/messagechannel/messageChannel.go
```diff
@@ -0,0 +1,61 @@
...
+	ctx context.Context,
+	sender syntax.DID,
+	receiver syntax.DID,
+	req *http.Request,
+) (*http.Response, error) {
+	atid, err := m.directory.LookupDID(ctx, receiver)
+	if err != nil {
+		return nil, fmt.Errorf("failed to lookup identity: %w", err)
+	}
+	pearServiceEndpoint := atid.GetServiceEndpoint("habitat")
```
arushibandi: nit: probably take in serviceName here now that we are separating habitat and habitat_local

internal/messagechannel/messageChannel.go
```diff
@@ -0,0 +1,61 @@
...
+	"github.com/habitat-network/habitat/internal/oauthclient"
+)
+
+type MessageChannel interface {
+	SendXRPC(
+		ctx context.Context,
+		sender syntax.DID,
+		receiver syntax.DID,
+		req *http.Request,
+	) (*http.Response, error)
```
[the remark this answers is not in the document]
arushibandi: ok idc about the name rn

internal/messagechannel/messageChannel.go
sashankg: will do after you merge yours

sashankg:
> i don't think this should be separate from pear

just trying to separate concerns so that we can swap out the implementation with something more robust. The intention of this interface is to abstract the actual message sending so we don't need to change pear to improve the message sending

PR #96, opened 2026-02-09
dependencies refactor

arushibandi:
renames/make things a little cleaner.

### DIFF
--- a/internal/inbox/inbox_test.go
+++ b/internal/inbox/inbox_test.go
@@ -11,8 +11,8 @@ import (
 	"gorm.io/gorm"
 )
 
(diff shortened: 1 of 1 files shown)


internal/inbox/inbox.go
```diff
@@ -32,7 +32,7 @@ type inbox struct {
 // inbox implements Inbox
 var _ Inbox = &inbox{}
 
-func NewInbox(db *gorm.DB) (Inbox, error) {
+func New(db *gorm.DB) (Inbox, error) {
```
sashankg: I don't like this. Doesn't hurt to be more explicit

PR #97, opened 2026-02-11
[ui]: fix logout button

arushibandi:


frontend/src/components/header.tsx
```diff
@@ -11,6 +12,22 @@ function formatHandle(handle: string | null) {
 }
 
 const Header = ({ authManager }: { authManager: AuthManager }) => {
+  const [isAuthenticated, setIsAuthenticated] = useState(
+    authManager.isAuthenticated()
+  );
+  const [handle, setHandle] = useState(authManager.handle);
+
+  useEffect(() => {
```
sashankg: This is not right. We can't react to the authmanager state so this effect won't be triggered when auth changes. The right way is to have an AuthenticatedHeader that is part of the /_requireAuth/ layout. In the body of the that you know that the user is signed in so you can just have a logout button that calls authmanager.logout

PR #99, opened 2026-02-11
[Dev] Use args instead of env for domain config

sashankg:
Apparently we can't use env vars within env vars so pass in the domains (which are derived from the TS_DOMAIN env var) via cli args

### DIFF
--- a/templates/web-app/moon.yml
+++ b/templates/web-app/moon.yml
@@ -1,13 +1,26 @@
 tasks:
   dev:
-    env:
-      DOMAIN: "$project.${TS_DOMAIN}"
+    args: 
+      - --
+      - --domain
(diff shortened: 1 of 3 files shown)


sashankg:
> can we change it to just DOMAIN?

what do you want to change to just `DOMAIN`? 

if you're talking about `TS_DOMAIN` then I think it should be explicit that it has to do with tailscale. why make it more ambiguous?

if you're talking about `HABITAT_DOMAIN` then no we automatically prefix all flag names with `HABITAT_` when reading them as env vars to avoid colliding with other env vars. 

if you're talking about the frontend, then that already is being read as `DOMAIN` which is probably bad and we should rename it to `FRONTEND_DOMAIN` to be more clear

PR #100, opened 2026-02-11
[Frontend] Implement logout

sashankg:


frontend/src/main.tsx
```diff
@@ -12,6 +12,7 @@ import reportWebVitals from "internal/reportWebVitals.ts";
 import { AuthManager } from "internal/authManager.ts";
 
 const authManager = new AuthManager(__DOMAIN__, __HABITAT_DOMAIN__, () => {
+  router.invalidate();
```
arushibandi: ooh what does this do?

sashankg: it invalidates all the data loaded in the "loaders" for each route. In our case we load the current auth state in the root loader so we invalidate it on logout

PR #104, opened 2026-02-12
[groups]: lexicons

arushibandi:


sashankg (commented):
why do groups need to be a primitive? isn't it the same as having a "group" record (like a regular record stored in the repo) and assigning it permissions based on its users. and then you can assign other records to that group record's arena?

arushibandi:
i think it gets confusing because i want the group to hold a list of members, but this list is == to the permissions it grants. so if you model this as a typical record, it would be possible for the grantees and the members list to get out of sync. and you also want to be able to fetch the list of members in a group; if you have a single source of truth for this, like if you make it the grantees list, then we're overloading permissions to get the members of a group out of it.

so i do think we will need a group primitive (and probably still the APIs i defined for convenience)

i think the way we will want to define this is that for a record, grantees can be either DIDs or habitat uris. if the habitat uri points to a group record, we use the members of the record for the permission check. otherwise, we just look at the record's grantees.

a group record is a special type of collection in the repo then -- its grantees are always == members, and thats enforced by the pear node.

sashankg:
> if you have a single source of truth for this, like if you make it the grantees list, then we're overloading permissions to get the members of a group out of it.

I disagree that it's overloading. For whatever group type we define, we need to sync its permissions with its member list, right? Either we define a special type of object where the user list is also the permission or we reuse the existing record object type and define its user list as the grantees. Both can be considered "overloading". I'm strongly in favor of not adding more things to the API beyond `put/get/listRecord`. Less API means less things that we need to commit to long term. We can paper over this with client libraries if we don't want to expose that to developers but we can figure that out later.

arushibandi:
> > if you have a single source of truth for this, like if you make it the grantees list, then we're overloading permissions to get the members of a group out of it.
> 
> I disagree that it's overloading. For whatever group type we define, we need to sync its permissions with its member list, right? Either we define a special type of object where the user list is also the permission or we reuse the existing record object type and define its user list as the grantees. Both can be considered "overloading". I'm strongly in favor of not adding more things to the API beyond `put/get/listRecord`. Less API means less things that we need to commit to long term. We can paper over this with client libraries if we don't want to expose that to developers but we can figure that out later.

But then what does `getGroupMembers` look like? that seems like a must-have api. it would need to be `getPermissions` on a given record, which seems weird no? and that would also be a new api beyond put/get/list

arushibandi:
another thing is i think it simplifies the model to pull this out as its own primitive. For example, with the calendar app: if you want to create a calendar for a group, you can have a cleaner structure by creating a group directly, then assigning a `calendar` lexicon to have that group as its sole grantee, then having that calendar hold a list of events, which each create their own clique. 

This is cleaner than making the group a clique, because then it would be some weird form of nesting.

PR #107, opened 2026-02-12
cleanup permission

arushibandi:
Actually fix the active: false thing (it was related to the fosite library writing a 200 on an error not the other thing I thought) and also fill in the permissions UI for the People tab (show permissions by DID). Also generate lexicons for permissions endpoints.

### DIFF
--- a/internal/oauthserver/oauth_server.go
+++ b/internal/oauthserver/oauth_server.go
@@ -350,6 +350,8 @@ func (o *OAuthServer) Validate(
 		scopes...,
 	)
 	if err != nil {
+		// TODO: we should delegate the response to o.provider.WriteIntrospectionError(ctx, w, err)
+		// Unfortunately that was returning a 200 http response, so we write our own error here.
 		utils.WriteHTTPError(w, fmt.Errorf("invalid or expired token: %w", err), http.StatusUnauthorized)
 		return "", false
 	}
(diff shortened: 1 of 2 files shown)


frontend/src/routes/_requireAuth/permissions/people/$did.tsx
```diff
@@ -0,0 +1,76 @@
...
+    const data = await context.queryClient.fetchQuery(
+      listPermissions(context.authManager),
+    );
+    // Find all lexicons this person has access to
+    const lexicons: string[] = [];
+    for (const [lexicon, dids] of Object.entries(data)) {
+      if (dids.includes(params.did)) {
+        lexicons.push(lexicon);
+      }
+    }
```
sashankg: seems like we should add a param to the `listPermissions` endpoint to support filtering by grantee. this shouldn't need to be done on the client

arushibandi: Ya agree.

internal/oauthserver/oauth_server.go
```diff
@@ -351,7 +350,7 @@ func (o *OAuthServer) Validate(
 		scopes...,
 	)
 	if err != nil {
-		o.provider.WriteIntrospectionError(ctx, w, err)
+		utils.WriteHTTPError(w, fmt.Errorf("invalid or expired token: %w", err), http.StatusUnauthorized)
```
sashankg: Can we add a todo. I think we should delegate this to fosite so that it can provide a more detailed error message. I'll look into why it was returning that weird response

PR #109, opened 2026-02-12
[cleanup]: delete unused lexicons

arushibandi:
Some cleanup before cliques (let's call it cliques not arenas to avoid confusion with other atproto stuff)
- delete all internal APIs except for notifyOfUpdate (listRecords + getRecords will also be called from node to node, but that reuses the public API)
- actually parse out grantees in putRecord as either did's or cliques and reject cliques for now as its unsupported.
- turned out we weren't deleting old go generated files, so i updated the lexgen script to do that. unfortunately we were using old types that got deleted.


internal/pear/repo.go
```diff
@@ -277,3 +282,30 @@ func (r *repo) listRecords(
 	}
 	return rows, nil
 }
+
+// get all items in the given clique
+func (r *repo) getCliqueItems(cliqueURI string) ([]string, error) {
+	rows, err := gorm.G[CliqueItem](
```
sashankg: clique items are just regular record, no? why do we have to store them separately?

sashankg: if it's an index then shouldn't it be a literal index on the permissions table? like in order to fetch all the items in a clique, we just need to fetch permissions by `grantee==cliqueRkey` and something similar for the inbox to get other user records

PR #113, opened 2026-02-16
[cliques]: implementation

arushibandi:


internal/permissions/store.go
```diff
@@ -214,3 +218,23 @@ func (s *store) ListReadPermissionsByUser(
 
 	return allows, denies, nil
 }
+
+func (s *store) ListAllowedRecordsByGrantee(ctx context.Context, caller string, grantee string) ([]habitat_syntax.HabitatURI, error) {
+	permissions, err := gorm.G[Permission](s.db.Debug()).
```
sashankg: remove `Debug`

internal/repo/repo.go
```diff
@@ -83,20 +96,20 @@ func (r *repo) putRecord(did string, collection string, rkey string, rec map[str
...
 )
 
-func (r *repo) getRecord(did string, collection string, rkey string) (*Record, error) {
+func (r *repo) GetRecord(did string, collection string, rkey string) (*Record, error) {
 	// Query using separate collection and rkey fields
 	row, err := gorm.G[Record](
 		r.db,
 	).Where("did = ? AND collection = ? AND rkey = ?", did, collection, rkey).
-		First(context.Background())
+		First(r.ctx)
```
sashankg: this should be an arg. we need the request context here

internal/pear/pear.go
```diff
@@ -155,14 +168,95 @@ func (p *Pear) getBlob(
...
+
+	// Otherwise, ensure that this item is indexed by the clique owner.
+	// If the clique + item owners are the same, no need to take action since the item is indexed by way of the clique repo.
+	if did == cliqueDID {
+		// Nothing to do, return.
+		return nil
+	}
+
+	// Otherwise, ensure that that the item is discoverable by the clique owner via the inbox.
+	p.notifyOfUpdate(ctx, did, cliqueDID, collection.String(), rkey.String(), (*string)(&cliqueURI))
```
sashankg: no we shouldn't use the inbox for intra node communication. users on the same node use the same permission store and repo and so they don't need to be notified

internal/pear/pear.go
```diff
@@ -38,49 +49,109 @@ var (
...
+	)
+	if err != nil {
+		return "", err
+	}
+
+	if len(cliqueGrantees) == 0 {
+		return "", nil
+	}
+
+	// If we granted permission to a clique, we need to notify the clique owner.
```
sashankg: we need to notify the clique owner under all circumstances. not just if we grant permission

internal/pear/pear.go
```diff
@@ -129,40 +202,236 @@ func (p *Pear) listRecords(
...
+		return err
+	}
+
+	did, collection, rkey, err := itemURI.ExtractParts()
+	if err != nil {
+		return err
+	}
+
+	// If the clique + item owners are the same, no need to take action since the item is indexed by way of the clique repo.
+	if did == cliqueDID {
```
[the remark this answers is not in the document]
sashankg: alternatively, don't do this check and let `notifyOfUpdate` skip self notifications

internal/pear/pear.go
```diff
@@ -129,40 +202,236 @@ func (p *Pear) listRecords(
...
-) (string /* mimetype */, []byte /* raw blob */, error) {
-	return p.repo.getBlob(did, cid)
+// Clique-related methods
+func (p *Pear) addCliqueItem(ctx context.Context, cliqueURI habitat_syntax.HabitatURI, itemURI habitat_syntax.HabitatURI) error {
+	cliqueDID, err := cliqueURI.Authority().AsDID()
+	if err != nil {
+		return err
+	}
+
+	did, collection, rkey, err := itemURI.ExtractParts()
```
sashankg: can we call this `itemOwnerDid`

PR #123, opened 2026-02-20
[cliques]: implementation first pass

arushibandi:
`getRecord` + `putRecord` works for cliques.

### DIFF
--- a/internal/pear/pear.go
+++ b/internal/pear/pear.go
@@ -23,7 +23,7 @@ import (
 //
 // This is the core of the habitat server.
 type Pear interface {
-	permissions.Enforcer
+	permissions.Store
 
 	// Permissioned repository methods
 	PutRecord(ctx context.Context, callerDID, targetDID syntax.DID, collection syntax.NSID, record map[string]any, rkey syntax.RecordKey, validate *bool, grantees []permissions.Grantee) (habitat_syntax.HabitatURI, error)
@@ -51,7 +51,7 @@ type pear struct {
 	xrpcCh xrpcchannel.XrpcChannel
 
 	// Backing for permissions
-	permissions permissions.Enforcer
+	permissions permissions.Store
 
 	// The backing store for the data. Should implement similar methods to public atproto repos
 	repo repo.Repo
@@ -61,13 +61,13 @@ type pear struct {
 }
 
 // Pass throughs to implement permission.Store
-func (p *pear) AddReadPermission(
+func (p *pear) AddPermissions(
 	grantees []permissions.Grantee,
 	owner syntax.DID,
 	collection syntax.NSID,
 	rkey syntax.RecordKey,
 ) error {
-	return p.permissions.AddReadPermission(grantees, owner, collection, rkey)
+	return p.permissions.AddPermissions(grantees, owner, collection, rkey)
 }
 
 // HasPermission implements Pear.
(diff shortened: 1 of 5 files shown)


internal/permissions/enforcer.go
```diff
@@ -0,0 +1,315 @@
...
+		rkey syntax.RecordKey,
+	) error
+	RemoveReadPermissions(
+		grantee []Grantee,
+		owner syntax.DID,
+		collection syntax.NSID,
+		rkey syntax.RecordKey,
+	) error
+	ListReadPermissions(
+		owner syntax.DID,
```
sashankg: I would personally swap the order of grantee and owner. grantee is the kind of the subject of this call and owner + collection + rkey is a combo of what they have permission to

internal/pear/pear.go
```diff
@@ -137,45 +174,126 @@ func (p *pear) PutRecord(
...
+	perms, err := p.permissions.ListReadPermissions(targetDID, callerDID, collection, rkey)
 	if err != nil {
 		return nil, err
 	}
 
-	if !authz {
-		return nil, ErrUnauthorized
+	// Try to find an explicit allow or deny permission
+	cliquesToResolve := []permissions.CliqueGrantee{}
+	for _, perm := range perms {
```
sashankg: i think `ListReadPermissions` when called with an `rkey` will only return 1 permission, right? if there's a specific permission for that record it will return that one. otherwise it will return a permission for the whole collection

internal/permissions/enforcer.go
```diff
@@ -0,0 +1,315 @@
...
+		query = query.Where("grantee = ? OR grantee LIKE ?", grantee.String(), "habitat://%") // check for the habitat uri prefix for cliques
+	}
+	if collection != "" {
+		query = query.Where("collection = ?", collection)
+	}
+	if rkey != "" {
+		// permissions with empty rkeys grant the entire collection
+		query = query.Where("rkey = ? OR rkey = ''", rkey)
+	}
+	query = query.Where("effect = ?", Allow)
```
sashankg: no we need to return the deny permissions here too, right? otherwise if a user has permission to a full collection but not for a specific record, the caller won't know that

internal/pear/pear.go
```diff
@@ -137,45 +174,126 @@ func (p *pear) PutRecord(
...
+		switch g := perm.Grantee.(type) {
+		case permissions.DIDGrantee:
```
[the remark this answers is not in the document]
arushibandi: that's what listreadpermissions does -- it returns both a direct permission (if it exists) + list of permissions on the record that could possibly give a grant via cliques

internal/pear/pear.go
arushibandi: The issue is you nthe issue is you need listreadpermissions to return anything that might grant permissions to that rkey, which includes cliques. we could split these up into two different methods on the enforcer to make that more clear (like hasPermission and CliquePermissions) but that's why it's written like that

arushibandi:
> can we try to reduce this PR to just the backend clique changes. I'm expecting there's gonna be some back and forth before we get it right and there's a lot of noise rn that might slow this down.

i'll do this for the future, but i'm gonna leave in the frontend changes for now.
i'm the only one who is keeping this up to date anyways, we keep pushing breaking changes. i made them mostly for my own testing sake to test the whole flow e2e and it worked

sashankg (approved):

PR #125, opened 2026-02-20
[permissions/node]: better abstractions

arushibandi:


internal/node/node.go
```diff
@@ -0,0 +1,14 @@
...
+import (
+	"context"
+
+	"github.com/bluesky-social/indigo/atproto/syntax"
+	"github.com/habitat-network/habitat/internal/xrpcchannel"
+)
+
+// An abstraction serving federated PDS ownership. Can be used by any component to answer the question: do I own the records/permissions/whatever for this DID?
+type Node interface {
+	xrpcchannel.XrpcChannel // TODO, should these merge into one type instead of embedding?
```
sashankg: why is the node responsible for sending messages? can't we just give the xrpc channel to whoever needs it?

internal/permissions/store.go
```diff
@@ -94,8 +103,50 @@ func NewStore(db *gorm.DB) (*store, error) {
...
+	case http.StatusUnauthorized, http.StatusForbidden:
+		return false, nil
+	default:
+		return false, fmt.Errorf("unexpected status from remote getRecord: %d", resp.StatusCode)
+	}
+}
+
 // HasPermission checks if a requester has permission to access a specific record.
-func (s *store) HasDirectPermission(
+func (s *store) HasPermission(
```
sashankg: this looks good to me

internal/server/server.go
```diff
@@ -354,34 +359,41 @@ func (s *Server) ListRecords(w http.ResponseWriter, r *http.Request) {
...
-	if !ok {
-		return
-	}
 
-	permissions, err := s.pear.ListPermissions("", callerDID, "", "")
-	if err != nil {
-		utils.LogAndHTTPError(w, err, "list permissions from store", http.StatusInternalServerError)
-		return
-	}
+	// TODO: fix later
```
sashankg: we can fix this now, right?

PR #126, opened 2026-02-21
TS typecheck

eagraf:
Noticed that stuff was breaking in various poitns of Typescript code,  turns out our internal package did not have a build script, and so our CI would not catch any type errors associated with it. Then, I fixed all the newly found issues when running CI.

Changes:
- Namespace path was wrong in `network.habitat.repo.listRecords`, causing the TS file to be misplaced when generated
- grantees missing from habitat client
- Remove notifications functions from habitat client
- Some other small client type fixes
·- remove permission: authenticated from lexicons

### DIFF
--- a/typescript/api/moon.yml
+++ b/typescript/api/moon.yml
@@ -8,7 +8,6 @@ fileGroups:
 workspace:
   inheritedTasks:
     exclude:
-    - build
     - start
     - dev
 tasks:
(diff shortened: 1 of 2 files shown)


sashankg (commented):
pnpm build  should be doing this already. if it's not for any project then we should add it to `pnpm build`

arushibandi (approved):

arushibandi (approved):

sashankg:
sorry i didn't see the updates

PR #129, opened 2026-02-21
[Repo] Fix putRecord on conflict

sashankg:


internal/repo/repo.go
```diff
@@ -69,33 +80,41 @@ func NewRepo(db *gorm.DB) (*repo, error) {
...
+			Where("collection = ?", collection).
+			Where("rkey = ?", rkey).
+			Delete(&Record{}).
+			Error; err != nil {
+			return err
+		}
+		bytes, err := json.Marshal(rec)
+		if err != nil {
+			return err
+		}
```
arushibandi: why do this inside the transaction?

PR #138, opened 2026-02-23
[AuthManager] Refresh tokens on expiration

sashankg:
forgot i actually need to implement this on the frontend

### DIFF
--- a/.moon/tasks/typescript-app.yml
+++ b/.moon/tasks/typescript-app.yml
@@ -27,5 +27,4 @@ tasks:
 
 inheritedBy:
   toolchain: typescript
-  language: typescript
   layer: application


typescript/apps/greensky/moon.yml
```diff
@@ -1,26 +1 @@
...
-      - --to
-      - $projectSource
-      - --force
-      - --
-      - --name
-      - $project
-    preset: utility
-    options:
-      runFromWorkspaceRoot: true
+layer: application
```
arushibandi: what does this mean?

typescript/internal/src/authManager.ts
```diff
@@ -4,18 +4,28 @@ import { decodeJwt } from "jose";
...
-const didLocalStorageKey = "did";
-const tokenLocalStorageKey = "token";
 const stateLocalStorageKey = "state";
 
-export class AuthManager {
-  handle: string | null;
-  did: string | null;
+interface AuthInfo {
+  did: string;
+  accessToken: string;
```
arushibandi: why not store handle anymore?

arushibandi (approved):

PR #139, opened 2026-02-24
[greensky]: public + private posts in a feed together -- demo

arushibandi:
sending a video in group because it's too big to upload here. got the greensky handle feed working with private posts. a bunch of other small changes to do this:
- listRecords supports the subjects param, to filter to some subset of users
- UI touch ups to greensky

Next steps:
- when creating a new post, allow posting 1) to your followers, 2) to only certain users, 3) to every one.

### DIFF
--- a/frontend/src/routes/_requireAuth/data.tsx
+++ b/frontend/src/routes/_requireAuth/data.tsx
@@ -4,7 +4,6 @@ import {
   useRouter,
 } from "@tanstack/react-router";
 import { useMemo } from "react";
-import { ListPrivateRecordsResponse } from "typescript/internal/src/habitatClient";
 
 interface SearchParams {
   lexicon?: string;


frontend/src/routes/_requireAuth/data.tsx
```diff
@@ -58,9 +58,21 @@ export const Route = createFileRoute("/_requireAuth/data")({
...
+        console.log("lexicon is", lexicon)
+        const response = await context.authManager.fetch(
+          "/xrpc/network.habitat.listRecords",
+          "POST",
+          JSON.stringify({
+            subjects: [repo],
+            collection: lexicon,
+          }),
+        );
+        const data: { records?: any[] } = await response.json();
```
sashankg: don't use `any` here. you should be able to get the actual response type from the api package

internal/server/server.go
```diff
@@ -300,20 +300,31 @@ func (s *Server) ListRecords(w http.ResponseWriter, r *http.Request) {
...
 		return
 	}
 
-	// TODO: fix this
-	if len(params.Subjects) > 0 {
-		utils.LogAndHTTPError(w, err, "don't allow filters by repo yet", http.StatusBadRequest)
-		return
+	dids := make([]syntax.DID, len(input.Subjects))
+	for i, subject := range input.Subjects {
+		// TODO: support handles
```
sashankg: why is this TODO? doesn't the below implementation support handles?

typescript/apps/greensky/src/routes/_requireAuth/handle.$handle.tsx
```diff
@@ -0,0 +1,118 @@
...
+            <li>
+              <Link to="/">← Greensky</Link>
+            </li>
+            <li>
+              <h2>@{handle}'s feed</h2>
+            </li>
+          </ul>
+          <ul>
+            <li>
+              <span>@{(authManager as any).handle}</span>
```
sashankg: don't do `as any`

sashankg (approved):

internal/permissions/store.go
```diff
@@ -404,7 +411,40 @@ func (s *store) RemovePermissions(
...
+		// Otherwise, it's a clique grantee, so we need to resolve it
+		// TODO: we could potentially be more efficient with the DB query than resolving each independently.
+		ok, err = s.isCliqueMember(ctx, grantee, clique)
+		if err != nil {
+			return nil, err
+		}
+
+		if ok {
+			// Keep all other fields of the permission the same
+			permission.Grantee = DIDGrantee(grantee)
```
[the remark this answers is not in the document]
arushibandi: We have to do this because the caller directly pipes the result of this into `ListRecords` which does not understand cliques.

PR #140, opened 2026-02-24
[greensky]: second part of demo

arushibandi:
Last PR was just public / private posts, and private posts needed permissions added manually.

This one the post modal has three options: public | followers | specific users. I also special cased a clique for every user which is keyed by `<did>/network.habitat.clique/followers`. That's the main change added from this PR. There's no sync-ing down from the PDS or anything, instead we just look up `getFollowers` from the public bsky endpoint to resolve this clique; it's special cased.

One thing to consider is in the future people could store all their data on habitat / we provide a PDS that is both public + private data. Then, the lookup of followers can't go through bsky because that would only index the public ones (we'd allow you to have private follows). So, we'd need to special case that, or change this clique to "public-bsky-follows" or something. I think more generally, this is a pattern where you want to create a clique and keep it up to date based on the existence of some other records that point to DIDs. Need to flesh out if we want to support this first class (i.e. you could imagine doing this for leaflet subscribers as well), or if we are only supporting it for bsky or something else.

Didn't take a video because i have 1000 tabs open right now but will do that tomorrow.

TODO: the public posting doesn't actually work because of formatting issues, will fix before merging but it's not really part of the demo.

### DIFF
--- a/internal/permissions/store.go
+++ b/internal/permissions/store.go
@@ -370,19 +370,21 @@ func (s *store) AddPermissions(
 		})
 	}
 
-	// Delete any existing permissions (allow or deny) for these grantees+record before inserting fresh allow permissions.
+	// Delete any existing permission for this record before inserting the deny.
 	// This is jank and its because SQLITE/postgres differ in the ON CONFLICT specs. We should fix this.
-	if err := s.db.Where("grantee IN ?", grantees).
-		Where("owner = ?", owner).
-		Where("collection = ?", collection).
-		Where("rkey = ?", rkey).
-		Delete(&permission{}).Error; err != nil {
-		return fmt.Errorf("failed to clear existing permissions before insert: %w", err)
-	}
-	if err := s.db.Create(&permissions).Error; err != nil {
-		return fmt.Errorf("failed to add lexicon permission: %w", err)
-	}
-	return nil
+	return s.db.Transaction(func(tx *gorm.DB) error {
+		if err := tx.Where("grantee IN ?", grantees).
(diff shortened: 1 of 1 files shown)


typescript/apps/greensky/src/routes/_requireAuth/index.tsx
```diff
@@ -120,15 +220,25 @@ export const Route = createFileRoute("/_requireAuth/")({
         <Feed entries={entries} />
         <dialog open={modalOpen}>
           <article>
-            <h1>New post</h1>
+            <header style={{ display: "flex", justifyContent: "space-between", alignItems: "center" }}>
```
sashankg: let's try to remove most of this custom styling. refer to https://picocss.com/docs for how to use the semantic tags to lay things out

typescript/apps/greensky/src/routes/_requireAuth/index.tsx
```diff
@@ -137,15 +247,94 @@ export const Route = createFileRoute("/_requireAuth/")({
...
+                          e.preventDefault();
+                          if (suggestions.length > 0) addUser(suggestions[0].handle);
+                          else if (userQuery.trim()) addUser(userQuery);
+                        }
+                      }}
+                      style={{ width: "100%" }}
+                    />
+                    {suggestions.length > 0 && (
+                      <ul style={{
+                        position: "absolute", top: "100%", left: 0, right: 0, zIndex: 10,
```
sashankg: same here remove all these custom styles

sashankg (approved):

PR #143, opened 2026-02-25
[ui]: better flow

arushibandi:
Moving everything we have into `/devtools` so I can start brainstorming + visualizing the new UI.
Also, updated it so that when you're not logged in it tells you to log in and doesn't show any thing else. When you're logged in but not onboarded to habitat, it redirects to the onboarding modal.

Screenshot 2026-02-24 at 6 00 49 PM


frontend/src/routes/index.tsx
```diff
@@ -1,88 +1,73 @@
...
 });
+
+function AuthenticatedHome({
+  authManager,
+}: {
+  authManager: ReturnType<typeof Route.useRouteContext>["authManager"];
+}) {
+  const did = authManager.getAuthInfo()!.did;
+
+  const { data: didDoc, isLoading } = useQuery({
```
sashankg: this should go in the loader

sashankg (approved):

PR #144, opened 2026-02-25
[greensky]: add to gh-pages

arushibandi:
- add greensky to gh pages so i can test in prod
- in the modal, if they don't have habitat in their did doc, prompt to add. i noticed i could still do writes because we don't actually gate the `PutRecord`, so i added a comment there


typescript/apps/greensky/src/routes/_requireAuth/NewPostButton.tsx
sashankg: also we should move these components out of the `routes/` directory and into a `components/` directory at the same level as `routes`

sashankg (approved):

PR #146, opened 2026-02-25
[bugfix]: never cancel the context passed into identity cache, except on shutdown.

arushibandi:
Context errors are cached. We see this specifically during an OAuth flow, where the browser re-directs a number of times, and then on the redirect back to habitat, on page load, we fail to load because a context cancelled error is cached. This type of error should not be cached. I opened a PR in my fork of indigo to demonstrate the issue: https://github.com/bluesky-social/indigo/pull/1345.

Seems like these should never be cached, but I'm also wondering if we're doing something wrong in our flow that is causing this. This is a temporary workaround (I see this like 30 times a day, on each login, which I'm doing frequently for testing reasons); either we should stop this from happening or merge the fix in the upstream repo (maybe both?)

### DIFF
--- a/internal/authn/service_auth.go
+++ b/internal/authn/service_auth.go
@@ -1,6 +1,7 @@
 package authn
 
 import (
+	"context"
 	"net/http"
 	"strings"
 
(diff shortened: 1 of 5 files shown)


sashankg (approved):
that's annoying. nice find!

PR #147, opened 2026-02-25
[pear]: add authz to all methods

arushibandi:
The permissions store is just a store -- it doesn't actually act on the authz that it encodes. The `pear` package is the "enforcing" part of "permissions enforcing atproto repository". It needs to do authz on every method.

There's a bit of duplication here where right now some of the methods on `pear` are only ever called by a certain caller from `server.go`. But this might not be true forever (i.e. if there's a new user of `pear`), so might as well gate all the methods now and enforce that pear methods have a `caller` attached for authz purposes.

I'm about to use this in the greensky demo to show different views in your feed depending on if a given post is public/followers only/specific users only.


sashankg (approved):
nice

PR #155, opened 2026-02-28
[Pear] Add self fallback

sashankg:


internal/node/node.go
```diff
@@ -23,27 +23,50 @@ type node struct {
...
 	}
+	for _, opt := range options {
+		opt(n)
+	}
+	return n
+}
+
+type NodeOption func(*node)
+
+func SelfFallback() NodeOption {
```
arushibandi: nit: I usually name options `With__` so `WithSelfFallback()`. it's a bit more readable.

arushibandi (approved):

PR #180, opened 2026-03-03
[Greensky] Pretty (1/n)

sashankg:
first pass at prettying things up. using [shadcn](https://ui.shadcn.com/) to get good baseline styles. the point of it is to copy the code into your codebase so that you can customize it yourself. for now sticking with the defaults. still trying not to add custom styles as much as possible and use whatever we get for free from the design system.

mainly focused on the new post modal and the post cards. will keep iterating 

Screenshot 2026-03-02 at 11 43 36 PM
Screenshot 2026-03-02 at 11 43 12 PM

### DIFF
--- a/typescript/internal/src/components/UserCombobox.tsx
+++ b/typescript/internal/src/components/UserCombobox.tsx
@@ -1,4 +1,4 @@
-import { AuthManager } from "@/authManager";
+import { AuthManager } from "../authManager";
 import {
   Combobox,
   ComboboxChip,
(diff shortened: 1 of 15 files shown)


typescript/apps/greensky/src/Feed.tsx
```diff
@@ -45,160 +61,110 @@ export function Feed({
...
+                    </ItemContent>
+                  </Item>
+                </ItemContent>
+                <ItemActions>
+                  <Badge variant="secondary">
+                    {entry.kind === "public"
+                      ? "🌍 Public"
+                      : entry.kind === "followers-only"
+                        ? "🔒 Followers only"
+                        : "👥 Specific users"}
```
arushibandi: cute

arushibandi (approved):

PR #181, opened 2026-03-04
[pear2p]: authn + authz in peer-discovery protocol for p2p

arushibandi:
Tested service auth by removing the oauth token from the request to peer-discovery and ensuring the peers are still connected. Did not test the malicious case where a peer does not have the right tokens and can't create p2p, should add a unit test for this.

Lots to do still to get it all tied up -- need to verify aud == habitat, create and publish a lexicon for this peer discovery and verify that.

Some open questions:
- should the peers be able to stay connected if the server goes offline? does that break permissioning if someone who loses permission can still sync to others with permission when the server is offline? and then they push those changes and everyone merges them in?

### DIFF
--- a/typescript/apps/docs/src/routes/_requireAuth/$uri.tsx
+++ b/typescript/apps/docs/src/routes/_requireAuth/$uri.tsx
@@ -2,7 +2,7 @@ import { Editor, EditorContent, useEditor } from "@tiptap/react";
 import StarterKit from "@tiptap/starter-kit";
 import { HabitatDoc } from "@/habitatDoc";
 import { useMutation } from "@tanstack/react-query";
-import { createFileRoute, redirect } from "@tanstack/react-router";
+import { createFileRoute } from "@tanstack/react-router";
 import { useMemo, useState } from "react";
 import { createLibp2p } from "libp2p";
 import { webSockets } from "@libp2p/websockets";
--- a/typescript/apps/docs/src/routes/_requireAuth/index.tsx
+++ b/typescript/apps/docs/src/routes/_requireAuth/index.tsx
@@ -39,7 +39,8 @@ export const Route = createFileRoute("/_requireAuth/")({
             },
           }),
         );
-        const { uri } = await response?.json();
+
+        const { uri } = await response.json();
         navigate({
           to: "/$uri",
           params: {


frontend/moon.yml
```diff
@@ -7,7 +7,7 @@ tasks:
       - --habitatDomain
       - pear.${TS_DOMAIN}
     deps:
-      - target: pear:dev
```
sashankg: Rm

typescript/apps/docs/src/routes/_requireAuth/$uri.tsx
```diff
@@ -101,12 +111,27 @@ export const Route = createFileRoute("/_requireAuth/$uri")({
...
+      const data = await response?.json();
+      return data.token;
+    }
+
+
     async function startPeerDiscovery(): Promise<void> {
       try {
         const stream = await node.dialProtocol(
           peerIdFromString(relayPeerId), "/habitat/peer-discovery/1.0.0");
+        const oauthToken = context.authManager.getAuthInfo()?.accessToken ?? "";
```
sashankg: let's just always do service auth, right? makes it simpler

internal/authn/service_auth.go
```diff
@@ -30,44 +30,58 @@ func (p *pdsServiceAuthMethod) CanHandle(r *http.Request) bool {
...
+	}
+	if err := token.Claims(atcryptoVerifier{publicKey}); err != nil {
+		return "", false, err
+	}
+	// TODO: validate Aud and Lxm
+	return issuerDid, true, nil
+}
+
+// Validate implements [Method].
+func (p *pdsServiceAuthMethod) Validate(
```
sashankg: every auth method's validate should just be calling its own ValidateRaw, right?

typescript/apps/docs/src/routes/_requireAuth/$uri.tsx
```diff
@@ -141,7 +166,11 @@ export const Route = createFileRoute("/_requireAuth/$uri")({
       Y.applyUpdateV2(ydoc, Uint8Array.fromBase64(data.value.blob));
     }
 
-    const did = context.authManager.getAuthInfo()?.did;
+    if (!context.authManager.getAuthInfo()) {
```
sashankg: shouldn't need to this. all routes in `_requireAuth/` already do

PR #183, opened 2026-03-04
[pear]: implement remote blob fetches

arushibandi:


internal/pear/pear.go
```diff
@@ -419,7 +422,53 @@ func (p *pear) ListRecords(ctx context.Context, caller syntax.DID, collection sy
...
+
+	resp, err := p.node.SendXRPC(ctx, caller, target, req)
+	if err != nil {
+		return "", nil, err
+	}
+	defer func() { _ = resp.Body.Close() }()
+
+	switch resp.StatusCode {
+	case http.StatusOK:
+		slurp, err := io.ReadAll(resp.Body)
```
sashankg: would be nice to just pipe this response through to the initiating request instead of loading it into memory

sashankg: should be easy, right? just return a reader

PR #197, opened 2026-03-09
[API] make putRecord rkey optional

sashankg:


sashankg:
accidentally changed the atproto putRecord here oops. fixed in https://github.com/habitat-network/habitat/pull/195

PR #198, opened 2026-03-09
[console]: v0 of data exploration

arushibandi:
In index.tsx, show a row of all the collections + metadata for them.
When you click on a collection card, it takes you to the exploration page for that collection, which for now shows record cards: truncated raw data for each record + who has access.

https://github.com/user-attachments/assets/c048395e-349f-4d6a-ae96-955606e613af


frontend/src/routes/_requireAuth.tsx
```diff
@@ -1,7 +1,8 @@
 import { createFileRoute, Outlet, redirect } from "@tanstack/react-router";
 
 export const Route = createFileRoute("/_requireAuth")({
-  beforeLoad({ context }) {
+  async beforeLoad({ context }) {
+    await context.authManager.maybeExchangeCode();
```
sashankg: this should just be called once in __root.tsx

arushibandi:
I've been thinking about it more... I do think the cards make sense for the homepage "preview", and i think for the records page, we'd need to render each type in a custom way, so it's going to depend. like bluesky posts maybe should be cards but docs should maybe be rows? i'm not sure yet

Either way, I'm going to merge what I have because it gets the idea across and continue working on the console

PR #200, opened 2026-03-10
[server]: don't log ErrUnauthorized

arushibandi:
We are seeing this in the server logs it's noisy.

### DIFF
--- a/cmd/pear/pds_forwarding.go
+++ b/cmd/pear/pds_forwarding.go
@@ -7,6 +7,7 @@ import (
 	"github.com/habitat-network/habitat/internal/oauthserver"
 	"github.com/habitat-network/habitat/internal/pdsclient"
 	"github.com/habitat-network/habitat/internal/pdscred"
+	"github.com/habitat-network/habitat/internal/utils"
 	"github.com/rs/zerolog/log"
 )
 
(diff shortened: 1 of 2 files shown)


cmd/pear/pds_forwarding.go
```diff
@@ -75,7 +74,9 @@ func (p *pdsForwarding) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 	// Copy response body
 	if _, err := io.Copy(w, resp.Body); err != nil {
 		// Log error but can't change status code at this point
-		log.Error().Err(err).Msg("failed to copy response body")
+		if utils.ShouldLog(err) {
```
sashankg: this won't catch unauthorized errors from the response, right?

sashankg: as in  `errors.Is(err, habitat_err.ErrUnauthorized)` won't return true

internal/utils/http.go
```diff
@@ -29,8 +30,8 @@ func LogAndHTTPError(w http.ResponseWriter, err error, debug string, code int) {
 	_ = json.NewEncoder(w).Encode(&ErrorMessage{Error: "unknown error"})
 }
 
-func shouldLog(err error) bool {
-	return !errors.Is(err, context.Canceled)
+func ShouldLog(err error) bool {
+	return !errors.Is(err, context.Canceled) && !errors.Is(err, habitat_err.ErrUnauthorized)
```
sashankg: we should probably add meters for this since we're not logging them. makes me think we need a more robust error framework than just `LogAndHttpError`

PR #201, opened 2026-03-10
[oauth server]: remove habitat service stuff

arushibandi:
Commenting this out for now because we don't require people to be onboarded.
Drive by change as I was going to add metrics.


sashankg (commented):
can we instead use the `Node` object and call ServesDID which will return true because of SelfFallback.

PR #202, opened 2026-03-10
[oauth]: couple of basic metrics

arushibandi:
@sashankg can you tell me if this is correct:

- rate of successful logins  = callback success / (callback err + authorize err + callback success)
- overall rate of failed logins = (callback err + auth err) / callback success
- overall rate of incompletes = 1 - (authorize success / callback success)

Also, do you think we should tag by anything? I was thinking possibly by user domain or `bsky.social` vs not `bsky.social`, but wondering if that's too much pii (like if there's just one handle for a domain we know exactly who it is). Just thinking in case we run into issues with only non-bsky domains or non-bsky PDS'es or vice versa for some reason.

### DIFF
--- a/internal/oauthserver/oauth_server_test.go
+++ b/internal/oauthserver/oauth_server_test.go
@@ -3,7 +3,6 @@ package oauthserver
 import (
 	"context"
 	"encoding/json"
-	"fmt"
 	"io"
 	"net/http"
 	"net/http/cookiejar"
@@ -11,7 +10,6 @@ import (
 	"testing"
 	"time"
 
-	"github.com/bluesky-social/indigo/atproto/identity"
 	"github.com/bluesky-social/indigo/atproto/syntax"
 	"github.com/gorilla/securecookie"
 	"github.com/gorilla/sessions"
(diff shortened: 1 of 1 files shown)


internal/oauthserver/oauth_server.go
```diff
@@ -106,7 +194,15 @@ func NewOAuthServer(
...
 	gob.Register(&authRequestFlash{})
 	gob.Register(pdsclient.AuthorizeState{})
+
+	oauthMetrics, err := newMetrics(meter)
+	if err != nil {
+		return nil, err
+	}
+
 	return &OAuthServer{
+		meter:           meter,
```
sashankg: do we need this?

internal/oauthserver/oauth_server.go
```diff
@@ -147,32 +251,38 @@ func (o *OAuthServer) HandleAuthorize(
...
 	}
 	if r.ParseForm() != nil {
+		o.metrics.authorizeErr(err)
 		utils.LogAndHTTPError(w, err, "failed to parse form", http.StatusBadRequest)
 		return
 	}
 	handle := r.Form.Get("handle")
 	atid, err := syntax.ParseAtIdentifier(handle)
 	if err != nil {
+		o.metrics.authorizeErr(err)
```
sashankg: same here and everywhere. if there are different reasons for the failed parse, should those be tracked indiviually or do we just need "bad handle"

PR #251, opened 2026-04-03
[permissions]: remove collection-level grants

arushibandi:
Found a bug in docs where if you have a collection-level grant but are not a member of the clique for that doc, its unclear what should happen.

Generally, this is going to be confusing with cliques, and seems like we likely don't need it. Let's delete it and we can add it back later if we need to.

### DIFF
--- a/internal/pear/pear.go
+++ b/internal/pear/pear.go
@@ -373,19 +373,31 @@ func (p *pear) listRecordsLocal(
 		return nil, fmt.Errorf("only support filtering by a collection")
 	}
 
+	fmt.Println("listrecordslocal", collection, caller, subjects)
 	// Resolve grants for both the caller and the clique
 	perms, err := p.permissions.ListGranteePermissions(ctx, caller, collection, subjects)
 	if err != nil {
 		return nil, err
 	}
 
-	records, err := p.repo.ListRecords(ctx, perms)
+	// Exclude caller's own records from the permission-based query — those are fetched
+	// separately below, and a caller may have granted their own records to a clique they belong to.
+	otherPerms := xslices.Filter(perms, func(p permissions.Permission) bool {
+		return p.Owner != caller
+	})
+
+	permissioned, err := p.repo.ListRecordsFromPermissions(ctx, otherPerms)
 	if err != nil {
-		return nil, fmt.Errorf("failed to list records: %w", err)
+		return nil, fmt.Errorf("failed to list permissioned records: %w", err)
+	}
+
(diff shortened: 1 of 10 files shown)


internal/pear/pear.go
```diff
@@ -374,18 +374,31 @@ func (p *pear) listRecordsLocal(
 		return nil, fmt.Errorf("only support filtering by a collection")
 	}
 
-	perms, err := p.permissions.ResolvePermissionsForCollection(ctx, caller, collection, subjects)
+	fmt.Println("listrecordslocal", collection, caller, subjects)
```
sashankg: rm

internal/permissions/store.go
```diff
@@ -111,6 +106,19 @@ func NewStore(db *gorm.DB, cliqueStore clique.Store) (*store, error) {
 		return nil, fmt.Errorf("failed to migrate permissions table: %w", err)
 	}
 
+	// Drop the effect column if it still exists
+	if db.Migrator().HasColumn(&permission{}, "effect") {
```
sashankg: we should use https://github.com/pressly/goose to properly document migrations. especially if we expect there to be multiple instances. this Migrator approach seems like it will get unwieldy quick. If you don't wanna set it up I would just not do this at all/do it manually

internal/permissions/store.go
```diff
@@ -127,26 +135,37 @@ func (s *store) HasPermission(
...
 
-	permissions, err := s.listPermissions(requester, []syntax.DID{owner}, collection, rkey)
+	if rkey == "" {
+		return false, ErrCollectionLevelNotSupported
+	}
+
+	// TODO: this query shouldn't use listPermissions
+	var permissions []permission
+	err := s.db.
+		Where("grantee = ? OR grantee LIKE ?", requester.String(), "clique:%"). // check for the habitat uri prefix for cliques
```
sashankg, line 131: we should really figure out how to do joins here so we can find all the permissions for cliques the caller is part of in one go

sashankg (approved):

internal/permissions/store.go
arushibandi: yeah, we might wanna move the clique package into permissions to do this

internal/permissions/store.go
arushibandi: ok i will just do it manually for now. this package looks nice, good find

arushibandi:
documenting my order of operations here:
--> remove rows where rkey="" (there weren't actually any)
--> deploy this code that stops using effect
--> once the code is out, drop the effect column in the permissions table

PR #253, opened 2026-04-03
[auth]: fix logouts

arushibandi:
The `(!this.refreshPromise)` only works in a single tab. This means if you have two tabs running habitat, they can race with each other to refresh tokens and logout.

Also small bug fix in `nonce_provider.go`.


sashankg (approved):
i guess this is good but this is definitely not what the actual is, right? we were getting signed out with just one tab open

arushibandi:
> i guess this is good but this is definitely not what the actual is, right? we were getting signed out with just one tab open

i've seen it happen with both : /

PR #254, opened 2026-04-03
[docs]: sidebar fixes

arushibandi:
It was pretty buggy, so refactor it to always pull from the latest ydoc state and persist ydoc state across doc tab switches.


sashankg (approved):
this seems very very wrong to me but idk how invested we are in maintaining this version of the app

PR #255, opened 2026-04-04
[dev]: postgres in local

arushibandi:
We should just use postgres in local. This means we need to run docker but oh well. It's more annoying to write code that works with both SQLite and Postgres (for ex. they handle ON CONFLICT differently), and FTS isn't supported in sqlite, so we need a way to test it. 

The first time you run `moon r :dev` it will fail because it's pulling the docker image so server startup fails to connect to postgres, but after that it should work.

### DIFF
--- a/README.md
+++ b/README.md
@@ -17,6 +17,8 @@ We use [moonrepo](https://moonrepo.dev/docs) to manage our monorepo. A crash cou
 
 `moon :dev` tasks read environment variables from `dev.env` which is gitignored.
 
+Add `HABITAT_PGURL=postgresql://habitat:habitat@localhost:5432/habitat` to your dev.env in order to connect to postgres and not sqlite.
+
 ### Tailscale Funnel
 
 We use Tailscale Funnel to make local services reachable from the public internet. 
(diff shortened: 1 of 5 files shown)


infra/moon.yml
sashankg, line 1: i don't think this should be a project itself. We just need a docker file and in the pear:dev task we can run a command for `docker compose -f dev-compose.yml up` or whatever the command it

README.md
```diff
@@ -17,6 +17,8 @@ We use [moonrepo](https://moonrepo.dev/docs) to manage our monorepo. A crash cou
 
 `moon :dev` tasks read environment variables from `dev.env` which is gitignored.
 
+Add `HABITAT_PGURL=postgresql://habitat:habitat@localhost:5432/habitat` to your dev.env in order to connect to postgres and not sqlite.
```
sashankg: we can just hard code the env in the pear:dev task

arushibandi:
@sashankg addressed your comments

sashankg (approved):

PR #257, opened 2026-04-07
[sql]: do proper on conflict now

arushibandi:
use on conflict instead of deletes + creates in a transaction

### DIFF
--- a/internal/clique/clique.go
+++ b/internal/clique/clique.go
@@ -178,10 +178,7 @@ func (s *store) RemoveMembers(clique habitat_syntax.Clique, members []syntax.DID
 		}
 	})
 
-	return s.db.Transaction(func(tx *gorm.DB) error {
-		// If that passes, try creating the clique members (no-op if exists already)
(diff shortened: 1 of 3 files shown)


sashankg (approved):
how are we gonna testing this then? If the current CI is passing then I guess we weren't testing collisions? I also would like to keep compatibility with sqlite since that will probably easier to host/spin up instances with. 

Also if we're doing this, we should remove/comment out the sqlite init stuff in pear/main.go

arushibandi:
> how are we gonna testing this then? If the current CI is passing then I guess we weren't testing collisions? I also would like to keep compatibility with sqlite since that will probably easier to host/spin up instances with.
> 
> Also if we're doing this, we should remove/comment out the sqlite init stuff in pear/main.go

The implementation for FTS looks really different between postgres and sqlite so i wasn't sure if we'd want to support both.

If we ship a docker image with postgres, i don't see how sqlite is necessarily easier?

PR #259, opened 2026-04-07
[frontend]: record renderer

arushibandi:
basic render schema + better rendering for records.


lexicons/network/habitat/render/schema.json
sashankg, line 1: this is a little rudimentary but fine for now. the ultimate goal is for "title", "meta", and "body" you would be able to compose fields from the record. so something like
```
title: [{ ... key: "eventName" }, { ... key: "startTime" }]
```

that way, when it's rendered, the title of a calendar event in search would look like "Alice's birthday 05/10". 

You can introduce other functionality like prefixes/suffixes to achieve "Alice's birthday **on** 05/10", different date format types, icons, etc.

sashankg: you can probably simplify this further to just list the keys you want to render

sashankg (approved):
can you attach a screenshot of what it looks like

PR #265, opened 2026-04-09
[orgs]: APIs

arushibandi:
Basic types for orgs + APIs. Not hooked up to the server yet since I want to get alignment on this first.


internal/server/server.go
sashankg, line 1: we should move the org server to its own package

internal/org/org.go
sashankg, line 1: why is this not in `internal/org`

internal/org/store.go
```diff
@@ -0,0 +1,208 @@
...
+	"gorm.io/gorm/clause"
+)
+
+var (
+	ErrNotAdmin  = errors.New("caller is not an admin")
+	ErrLastAdmin = errors.New("org must have at least one admin")
+)
+
+// org.Store is used to manage the organization tied to a pear node
+// It enforces permissions on its methods; i.e non-admins cannot add admins or members.
```
sashankg, line 19: the org store should not enforce permissions

internal/org/store.go
```diff
@@ -0,0 +1,208 @@
...
+//
+// Eventually we can add management scopes to member vs. admin roles, to allow members to add other members etc.
+type Store interface {
+	// Function for one time use to bootstrap the admin who created the org after provisioning necessary infra.
+	bootstrapAdmin(ctx context.Context, bootstrapSecret string, admin syntax.DID) error
```
sashankg: it's weird to me to have an interface with non exported methods. I feel all of these can be exported

internal/server/server.go
```diff
@@ -32,9 +33,23 @@ type authMethods struct {
 	serviceAuth authn.Method
 }
 
+type options struct {
+	org org.Store
+}
+
+type Option func(*options)
+
+func WithOrg(org org.Store) Option {
```
sashankg: let's make this not an option. the public pear server will just be an org that includes everyone

internal/server/server.go
```diff
@@ -61,14 +83,37 @@ func NewServer(
 	return server
 }
 
+func (s *Server) authnWithOrg(w http.ResponseWriter, r *http.Request, authnMethod ...authn.Method) (syntax.DID, bool) {
+	callerDID, ok := s.authnWithOrg(w, r, authnMethod...)
```
sashankg: recursive?

sashankg: i think you meant to call auth.Validate here

internal/server/server.go
```diff
@@ -61,14 +83,37 @@ func NewServer(
...
+		return "", false
+	}
+
+	// Otherwise if this server doesn't belong to an org, return
+	if s.options.org == nil {
+		return callerDID, true
+	}
+
+	// Otherwise, only authn if part of org
+	ok, err := s.options.org.IsMember(r.Context(), callerDID)
```
sashankg: looks like we're not checking `ok` here?

sashankg (approved):

PR #266, opened 2026-04-09
[orgs]: lexicons + add to servers

arushibandi:
Add the org APIs to the habitat lexicons. Implement them in the server. In main add a boolean flag indicating whether this is for an org or not. If not, same path as previously. If so, add additional middleware blocking requests from non-org members.

Test Plan:
- for this PR, I'm just going ensure that our server continues to work as normal.
- after it merges, i will set up an org server just for us so we can start testing this out.
- then i'll work on the async provisioner


cmd/pear/flags.go
```diff
@@ -77,6 +78,12 @@ func getFlags() ([]cli.Flag, []cli.MutuallyExclusiveFlags) {
 				Required: true,
 				Sources:  getSources(fOauthClientSecret),
 			},
+			// TODO: should this be more than just a bool flag?
+			&cli.BoolFlag{
+				Name:     fOrg,
+				Usage:    "Whether the server is being run as part of an organization",
```
sashankg, line 84: i thought we decided there's no difference except the public one is an org that includes everybody. let's just ignore the public case for now and just build it for orgs, right?

sashankg (approved):
lexicons lgtm

cmd/pear/flags.go
arushibandi: i think we still need some indicator saying either its the public org or something else in main. i'd rather support both paths now. we can have an additional deployment of an org (the non-public one) for us

PR #267, opened 2026-04-13
[docs]: support open docs (for logged in users)

arushibandi:
Allow for open docs. this works the same except writing to public PDS. 

Some changes:
--> fix bug where getRecord from public PDS always hits the caller's PDS shard, not necessarily the right one
--> p2p server allows peer-discovery on arbitrary records

Follow ups:
--> re-discovery: when you open a public doc, we need to save some state so your sidebar shows it next time. for now we can add a section to the sidebar "Recently opened public" or something. Immediately write a doc when you open one.
--> for non logged-in users: should show a pop-up to enter a screen name. once that's selected, they can connect to the p2p network and edit like any other person. only difference is their edits get saved to local storage / indexed db (not sure what framework to use here yet) rather than their personal PDS.


cmd/pear/main.go
```diff
@@ -229,6 +218,10 @@ func run(_ context.Context, cmd *cli.Command) error {
 	mux.HandleFunc("/xrpc/network.habitat.clique.getMembers", pearServer.GetCliqueMembers)
 	mux.HandleFunc("/xrpc/network.habitat.clique.isMember", pearServer.IsCliqueMember)
 
+	// special-case: handle public getRecord explicitly since forwarding needs to go to the right host
+	// TODO: do this for listRecords / other getters
+	mux.HandleFunc("/xrpc/com.atproto.repo.getRecord", pearServer.GetPublicRecord)
```
sashankg, line 223: i thought `com.atproto.sync.getRecord` would resolve the right repo but that returns the car instead of `value`. I think the right way to do this would be to implement an appview for docs that indexes the public documents. 

in the meantime, let's just figure out the right repo on the client instead of messing with the semantics of the atproto API

lexicons/network/habitat/docs.json
```diff
@@ -25,6 +25,10 @@
                         "type": "string",
                         "format": "uri",
                         "description": "URI of the clique whose members may edit this document."
+                    },
+                    "isPublic": {
```
sashankg: we would know this based on the uri, right?

typescript/internal/src/habitatClient.ts
```diff
@@ -219,6 +231,39 @@ interface ListRecordsResponse<T extends Record<string, unknown>>
   records: TypedRecord<T>[];
 }
 
+export const getPublicRecord = async <T = Record<string, unknown>>(
+  authManager: AuthManager,
```
sashankg, line 235: you don't need to fetch public records via the auth manager. let's just resolve their pds and directly make a getRecord request there

typescript/internal/src/components/ShareDialog.tsx
```diff
@@ -6,61 +6,111 @@ import { AuthManager } from "@/authManager";
...
+            </div>
+            {publicUrl && (
+              <div className="flex items-center gap-2">
+                <input
+                  readOnly
+                  value={publicUrl}
+                  className="flex-1 truncate rounded border px-2 py-1 text-sm font-mono bg-muted"
+                />
+                <Button variant="outline" size="icon" onClick={handleCopy} aria-label="Copy link">
+                  {copied ? <CheckIcon className="size-4" /> : <CopyIcon className="size-4" />}
```
sashankg, line 64: use https://ui.shadcn.com/docs/components/base/sonner to show a temporary message

typescript/apps/docs/src/queries/docs.tsx
```diff
@@ -190,3 +237,24 @@ export const removePermissionMutationOptions = (authManager: AuthManager) =>
...
+        {
+          repo,
+          collection: "network.habitat.docs",
+          rkey,
+          record,
+        },
+        { authManager },
+      );
+      await deleteDoc(authManager, uri);
+      return `at://${repo}/network.habitat.docs/${rkey}`;
```
sashankg, line 258: use the uri returned from `com.atproto.repo.putRecord`

typescript/apps/docs/src/queries/docs.tsx
```diff
@@ -1,36 +1,74 @@
...
     queryKey: ["docs"],
     staleTime: 1000 * 60 * 5,
-    queryFn: () =>
-      listPrivateRecords<HabitatDoc>(authManager, "network.habitat.docs"),
+    queryFn: async () => {
+      const did = authManager.getAuthInfo()?.did;
+      const [privateResult, publicResult] = await Promise.all([
+        listPrivateRecords<HabitatDoc>(authManager, "network.habitat.docs"),
+        did
+          ? listPublicRecords<HabitatDoc>(authManager, "network.habitat.docs", did)
```
sashankg: do we wanna fetch this user's public edits too?

sashankg (commented):
a simpler "make public" would be to just publish a site.standard.document for others to just read it instead of making it editable by all

typescript/internal/src/habitatClient.ts
arushibandi: add a new lexicon that proxies through.

arushibandi:
comment so i don't forget: link from habitat/ to docs/ is broken

PR #268, opened 2026-04-15
[pear]: docker image for debian / collective action school

arushibandi:
Add docker containers for both backend and frontend (habitat home hub only)

Couple of bug fixes.

Follow-ups:
- Get org management server running
- Figure out what to do about oauth/ paths for orgs
- API docs clean-ups


frontend/index.html
sashankg, line 1: idk if the formatter complains about html but let's fix the indentation here

frontend/public/manifest.json
```diff
@@ -1,6 +1,6 @@
 {
-  "short_name": "TanStack App",
-  "name": "Create TanStack App Sample",
+  "short_name": "habitat",
+  "name": "habitat hoje",
```
sashankg: hoje?

cmd/pear/pds_forwarding.go
```diff
@@ -47,8 +47,17 @@ func (p *pdsForwarding) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 		utils.LogAndHTTPError(w, err, "[pds forwarding]: failed to create forwarding request", http.StatusInternalServerError)
 		return
 	}
-	// Copy headers from original request
+	// Copy headers from original request, stripping hop-by-hop headers that
+	// must not be forwarded (e.g. Connection: upgrade, which HTTP/2 rejects).
 	req.Header = r.Header.Clone()
+	for _, h := range strings.Split(req.Header.Get("Connection"), ",") {
```
sashankg, line 53: i think we should be entirely filtering these requests out, right? not modifying them

build/debian/pear/docker-entrypoint.sh
sashankg, line 1: ooh nice

sashankg (approved):

cmd/pear/pds_forwarding.go
arushibandi: I think proxies are generally supposed to include all headers except for hop-by-hop headers

PR #271, opened 2026-04-16
[orgs]: setup management server

arushibandi:
Set up management APIs + server


cmd/pear/main.go
```diff
@@ -153,27 +153,36 @@ func run(_ context.Context, cmd *cli.Command) error {
...
-		orgStore, err = org.NewStore(org.Org{Domain: domain}, db)
+		orgServer, err := org.NewServer(db, oauthServer)
 		if err != nil {
-			log.Fatal().Err(err).Msgf("unable to setup backing store for org with domain: %s", domain)
+			log.Fatal().Err(err).Msgf("unable to setup org server for domain: %s", domain)
 		}
+		orgMembership = orgServer
+
+		// org management routes — only available on org-serving nodes
+		mux.HandleFunc("/xrpc/network.habitat.org.getAdmins", orgServer.GetAdmins)
```
sashankg: we don't need to conditionally add these routes, right? no one is an admin of the everyone org so they'll be protected. we should really avoid forking behavior as much as possible and just keep the differences to the org.Org implementation

cmd/pear/main.go
```diff
@@ -153,27 +153,36 @@ func run(_ context.Context, cmd *cli.Command) error {
...
+
 	// TODO: take in non-everything org depending on CLI flag
 	servingOrg := cmd.Bool(fOrg)
 
 	// Default: no org == org that serves everyone
-	orgStore := org.NewEveryoneOrg()
+	var orgMembership org.Org = org.NewEveryoneOrg()
 	if servingOrg {
-		orgStore, err = org.NewStore(org.Org{Domain: domain}, db)
+		orgServer, err := org.NewServer(db, oauthServer)
```
sashankg: NewServer should not be returning an org.Org. It should return an org.Server and take an org.Org as input

internal/org/store.go
```diff
@@ -19,7 +19,7 @@ var (
 // It enforces permissions on its methods; i.e non-admins cannot add admins or members.
 //
 // Eventually we can add management scopes to member vs. admin roles, to allow members to add other members etc.
-type Store interface {
+type Org interface {
```
sashankg: yeah this is change for the worse imo

cmd/pear/main.go
arushibandi: I wanted this as well but i want a single org.Org i can pass around (to the pear server) that only implements IsMember

if i create the store in main, then i need to export all the other methods of org (or assert the type) so the server knows it has access to the other types

this feels harder than it should, i feel like somewhere i went wrong with this

arushibandi: it's also redundant for the org server to take in an org.Org (which must have a db in the non-everything case) and create its own store

arushibandi: NVM i fixed this

internal/org/store.go
sashankg, line 1: nit: rename this file and the test file

.testcoverage.yml
```diff
@@ -26,6 +26,7 @@ exclude:
     - ^internal\/.*\/server.go # Most of our servers are mostly pass throughs -- TODO, we should test them.
     - ^internal/telemetry # This is all setup, no logic
     - ^internal/utils
+    - internal/org/public.go # This is all boilerplate
```
sashankg: imo let's just add super simple tests for this. can't hurt. ai can do it real quick

sashankg (approved):

PR #272, opened 2026-04-17
[frontend]: first pass at admin UI

arushibandi:
https://github.com/user-attachments/assets/58d78759-d017-4aac-b2c4-c67b3b43d260


cmd/pear/main.go
```diff
@@ -175,13 +176,17 @@ func run(_ context.Context, cmd *cli.Command) error {
 		log.Fatal().Err(err).Msgf("unable to setup org server for domain: %s", domain)
 	}
 
+	// public config endpoint
+	mux.HandleFunc("/xrpc/network.habitat.getConfig", serveConfig(servingOrg, domain))
```
sashankg: if this is supposed to be public let's call that out in the endpoint name: `network.habitat.getPublicConfig` or something like that

cmd/pear/main.go
```diff
@@ -302,6 +307,23 @@ func run(_ context.Context, cmd *cli.Command) error {
 	return err
 }
 
+func serveConfig(servingOrg bool, domain string) http.HandlerFunc {
```
sashankg: let's do this in a separate file

sashankg: i can imagine this could also belong in the org store or at least the org server

frontend/src/routes/_requireAuth/org/index.tsx
```diff
@@ -0,0 +1,151 @@
...
+      <h2 className="text-lg font-semibold mb-2">{title}</h2>
+      <table className="w-full text-sm">
+        <tbody>
+          {dids.map((did) => (
+            <tr key={did} className="border-b">
+              <td className="py-2 pr-4 font-mono">{did}</td>
+              {isAdmin && (
+                <td className="py-2 flex gap-2 justify-end">
+                  {canPromote && onPromote && (
+                    <button
```
sashankg: use `internal/components/ui`

frontend/src/routes/_requireAuth/org/index.tsx
```diff
@@ -0,0 +1,151 @@
...
+                  </button>
+                </td>
+              )}
+            </tr>
+          ))}
+        </tbody>
+      </table>
+      {isAdmin && (
+        <div className="flex gap-2 mt-3">
+          <input
```
sashankg: same use design system input

frontend/src/routes/__root.tsx
```diff
@@ -15,25 +16,29 @@ export const Route = createRootRouteWithContext<RouterContext>()({
...
   },
   staleTime: 1000 * 60 * 60,
   component() {
     const { authManager } = Route.useRouteContext();
-    const { profile } = Route.useLoaderData();
+    const { profile, orgDomain } = Route.useLoaderData();
     return (
       <div className="flex flex-col items-center w-full justify-stretch gap-4">
-        <Header profile={profile} onLogout={authManager.logout} />
+        <Header profile={profile} orgDomain={orgDomain ?? undefined} onLogout={authManager.logout} />
```
sashankg: `?? undefined` is redundant

internal/org/server.go
```diff
@@ -220,6 +221,41 @@ func (s *Server) RemoveAdmin(w http.ResponseWriter, r *http.Request) {
 	}
 }
 
+func (s *Server) DowngradeAdmin(w http.ResponseWriter, r *http.Request) {
+	caller, ok := s.authnWithOrg(w, r, s.oauth)
+	if !ok {
+		return
+	}
+
+	var req struct {
```
sashankg: this should be from the api

frontend/src/queries/org.ts
sashankg, line 1: this should go in frontend/queries. internal is only for things shared by all apps

frontend/src/routes/_requireAuth/org/index.tsx
```diff
@@ -0,0 +1,151 @@
...
+  onRemove: (did: string) => Promise<void>;
+  onAdd: (did: string) => Promise<void>;
+  addLabel: string;
+  onPromote?: (did: string) => Promise<void>;
+  canPromote: boolean;
+}) {
+  const [input, setInput] = useState("");
+
+  const { mutate: handleAdd, isPending: adding } = useMutation({
+    mutationFn: () => onAdd(input),
```
sashankg, line 88: just do the mutation directly here. there's no need to wrap this in a mutation that just invokes another mutation in the parent

sashankg (approved):

frontend/src/routes/_requireAuth/org/index.tsx
arushibandi: talked offline

cmd/pear/main.go
arushibandi: removed the generic public config in favor of an org one. right now it is incomplete, just returns the domain. eventually this should be some metadata about the org.

PR #273, opened 2026-04-17
[orgs]: oauth server allowlists org members

arushibandi:
https://github.com/user-attachments/assets/d60d892e-d581-4da9-a30c-37ec2666e97c

### DIFF
--- a/internal/oauthserver/oauth_server.go
+++ b/internal/oauthserver/oauth_server.go
@@ -375,19 +375,6 @@ func (o *OAuthServer) HandleCallback(
 	}
 
 	did := arf.Did
-	// If there is a DID allowlist, check this against it
-	ok, err = o.allowlistFn(r.Context(), did)
-	if err != nil {
-		o.metrics.callbackErr(err, "allowlist_dids")
-		utils.LogAndHTTPError(w, err, "failed to lookup allowlist did", http.StatusInternalServerError)
-		return
-	}
-
-	if !ok {
-		o.metrics.callbackErr(err, "allowlist_dids")
-		utils.LogAndHTTPError(w, err, "user is not allowed to authenticate with this server", http.StatusUnauthorized)
-		return
-	}
 
 	recreatedRequest, err := http.NewRequest(http.MethodGet, "/?"+arf.Form.Encode(), nil)
 	if err != nil {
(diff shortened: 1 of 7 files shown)


internal/oauthserver/oauth_server.go
```diff
@@ -158,6 +158,9 @@ type OAuthServer struct {
 	oauthClient  pdsclient.PdsOAuthClient   // Client for communicating with AT Protocol services
 	directory    identity.Directory         // AT Protocol identity directory for handle resolution
 	storage      *store
+
+	// Allowlist of DIDs that can oauth with this server
+	allowlistFn func(context.Context, syntax.DID) (bool, error)
```
sashankg: can't the oauth server just take an org store and check `IsMember`. this seems overly general

internal/oauthserver/oauth_server.go
```diff
@@ -367,6 +373,9 @@ func (o *OAuthServer) HandleCallback(
 		utils.LogAndHTTPError(w, err, "failed to parse auth request flash", http.StatusBadRequest)
 		return
 	}
+
+	did := arf.Did
```
sashankg: just inline this wherever it's used

typescript/internal/src/authManager.ts
```diff
@@ -70,8 +70,16 @@ export class AuthManager {
...
 
   async maybeExchangeCode() {
-    const currentUrl = window.location.href;
-    const url = new URL(currentUrl);
+    const url = new URL(window.location.href);
+    const oauthError = url.searchParams.get("error");
+    if (oauthError) {
+      const description =
+        url.searchParams.get("error_description") ?? oauthError;
+      if (!window.location.pathname.includes("oauth-login")) {
```
sashankg, line 78: will this ever be false? the oauth server will redirect to whichever redirect_uri we specify so we know that it will always redirect to `/`

sashankg: we should also probably add an explicit `/callback` route instead of guessing (by checking the url params) if we should exchange the code like we do now

sashankg (approved):

internal/oauthserver/oauth_server.go
arushibandi: changed `org.Server` to take in generic `authn.Method`. can make this ... if it needs multiple

typescript/internal/src/authManager.ts
arushibandi: @sashankg is fixing this in a follow up PR to sepearate maybeExchangeCode into exchangeCode and refresh

PR #274, opened 2026-04-18
[Calendar] Server

sashankg:
Basic server that for now just helps the user auth with google


cmd/pear/moon.yml
```diff
@@ -20,7 +20,7 @@ tasks:
       - '8000'
       - pear
   postgres:
-    command: docker compose -f ../../infra/docker-compose.yml up
+    command: docker-compose -f ../../infra/docker-compose.yml up
```
arushibandi: interesting why did this work for me

cmd/calendarServer/server.go
```diff
@@ -0,0 +1,359 @@
...
+      "id": "#calendar",
+      "serviceEndpoint": "https://%s",
+      "type": "CalendarServer"
+    }
+  ]
+}`
+	w.Header().Set("Content-Type", "application/json")
+	fmt.Fprintf(w, template, s.domain, s.domain)
+}
+
```
arushibandi, line 87: we should factor this out into `utils` since we are using it in several places

typescript/apps/calendar/src/components/ConnectGoogleDialog.tsx
```diff
@@ -0,0 +1,58 @@
...
+  return (
+    <Dialog>
+      <DialogTrigger
+        render={<Button>Connect Google</Button>}
+        onClick={() => startConnect()}
+      />
+      <DialogContent>
+        <DialogTitle>Connect Google</DialogTitle>
+        <DialogDescription>
+          Connect your Google account to sync events with your calendar.
```
arushibandi, line 44: nit: make it clear in this description that this is a one-way sync (I.e. google events show up here but not vice versa, and you cannot edit google events here)

typescript/apps/calendar/src/collections/events.ts
```diff
@@ -1,9 +1,35 @@
...
+}
+
+interface GoogleCalendarEvent {
+  name: string;
+  startsAt?: string;
+  endsAt?: string;
+}
+
+const CALENDAR_SERVER_PROXY_HEADER =
+  "did:web:calendar-server.dwelf-mirzam.ts.net#calendar";
```
arushibandi, line 26: should this be hard coded?

cmd/calendarServer/server.go
```diff
@@ -0,0 +1,359 @@
...
+type CalendarEvent struct {
+	Name        string                  `json:"name"`
+	Description string                  `json:"description,omitempty"`
+	CreatedAt   string                  `json:"createdAt"`
+	StartsAt    string                  `json:"startsAt,omitempty"`
+	EndsAt      string                  `json:"endsAt,omitempty"`
+	Mode        string                  `json:"mode,omitempty"`
+	Status      string                  `json:"status"`
```
arushibandi, line 315: can we make Status typed string?

cmd/calendarServer/server.go
```diff
@@ -0,0 +1,359 @@
...
+	json.NewEncoder(w).Encode(ErrorOutput{
+		Error:   error,
+		Message: message,
+	})
+}
+
+type CalendarEvent struct {
+	Name        string                  `json:"name"`
+	Description string                  `json:"description,omitempty"`
+	CreatedAt   string                  `json:"createdAt"`
```
arushibandi, line 311: do we need a `Creator` for the owner? or i guess this is implicit from the uri?

cmd/calendarServer/server.go
```diff
@@ -0,0 +1,359 @@
...
+	StartsAt    string                  `json:"startsAt,omitempty"`
+	EndsAt      string                  `json:"endsAt,omitempty"`
+	Mode        string                  `json:"mode,omitempty"`
+	Status      string                  `json:"status"`
+	Locations   []CalendarEventLocation `json:"locations,omitempty"`
```
arushibandi, line 316: why is this and URIs lists?

cmd/calendarServer/server.go
```diff
@@ -0,0 +1,359 @@
...
+	session, err := s.store.GetSessionByDID(callerDID.String())
+	if err != nil {
+		writeError(w, "Unauthorized", "no valid session", http.StatusUnauthorized)
+	}
+	return session, true
+}
+
+func (s *Server) handleGetEvents(w http.ResponseWriter, r *http.Request) {
+	dump, _ := httputil.DumpRequest(r, true)
+	log.Printf("handle ConnectGoogle Request: %s", dump)
```
arushibandi, line 194: rm log

cmd/calendarServer/server.go
```diff
@@ -0,0 +1,359 @@
...
+	}
+
+	if session.GoogleAccessToken == "" {
+		w.Header().Set("Content-Type", "application/json")
+		json.NewEncoder(w).Encode(struct {
+			Events []CalendarEvent `json:"events"`
+		}{
+			Events: []CalendarEvent{},
+		})
+		return
```
arushibandi, line 214: ? don't we still have to fetch the other events?

cmd/calendarServer/server.go
```diff
@@ -0,0 +1,359 @@
...
+		log.Printf("Error updating tokens: %v", err)
+		http.Error(w, "failed to store tokens", http.StatusInternalServerError)
+		return
+	}
+
+	w.Header().Set("Content-Type", "application/json")
+	json.NewEncoder(w).Encode(map[string]string{
+		"message": "Google account connected successfully",
+	})
+}
```
arushibandi, line 177: where does this fetch the events so that when the user is done connecting they show up?

cmd/calendarServer/server.go
arushibandi: oh wait the server is only for google events right?

arushibandi (approved):

cmd/pear/moon.yml
sashankg: i think they both work. this was from when i was troubleshooting. will revert

typescript/apps/calendar/src/collections/events.ts
sashankg: yeah i think for now at least that's fine. this calendar app view isn't tied to your private data at all. it's just for public/google data

cmd/calendarServer/server.go
sashankg: the client will do that

cmd/calendarServer/server.go
sashankg: this is just what the google api returns

cmd/pear/moon.yml
sashankg: nvm `docker compose` didn't work for me. idk why

PR #280, opened 2026-04-21
[pear]: api cleanup

arushibandi:
This PR is best read commit-by-commit.

1. consistent naming (`network.habitat.repo.putRecord` instead of `network.habitat.putRecord`. same for permissions / cliques APIs)
2. move the clique server into the clique package for better modularity and to be consistent about what pear / org / cliques do
3. add `repo.createRecord` to be more consistent with the public endpoints
4. update frontends to use new paths
5. update PDS forwarding to allow-list certain paths so we know we aren't doing something incorrect / insecure here, the way `getRecord` was accidentally broken. Most of the frontend refactors are a result of this, since now endpoints should hit the public bluesky api.
6. Replace `repo.listCollections` with `repo.describeRepo` to be a bit more consistent with standard atproto

The remaining `com.atproto.repo.*` endpoints we don't implement are: `applyWrites` (batch writes), `importRepo` (import full public repo as a CAR file -- not sure if / whether we should do this), `listMissingBlobs` (see here: https://docs.bsky.app/docs/api/com-atproto-repo-apply-writes).

We also implement `getBlob` under `repo.getBlob` rather than `sync.getBlob` like the atproto version does. Not sure what to make of that. We don't implement any `com.atproto.server.*` or `com.atproto.sync.*` methods or any of the other sub-labels like identity, admin, etc.

There's a bunch of misc changes to greensky; this app is not being maintained so if they are breaking changes, whatever.

**Test Plan:**
- Made sure frontend + docs still worked.


cmd/pear/main.go
```diff
@@ -224,30 +232,32 @@ func run(_ context.Context, cmd *cli.Command) error {
...
+	mux.HandleFunc("/xrpc/network.habitat.repo.putRecord", pearServer.PutRecord)
+	mux.HandleFunc("/xrpc/network.habitat.repo.getRecord", pearServer.GetRecord)
+	mux.HandleFunc("/xrpc/network.habitat.repo.listRecords", pearServer.ListRecords)
+	mux.HandleFunc("/xrpc/network.habitat.repo.describeRepo", pearServer.DescribeRepo)
 	mux.HandleFunc("/xrpc/network.habitat.repo.deleteRecord", pearServer.DeleteRecord)
-
-	// blobs
-	mux.HandleFunc("/xrpc/network.habitat.uploadBlob", pearServer.UploadBlob)
-	mux.HandleFunc("/xrpc/network.habitat.getBlob", pearServer.GetBlob)
+	mux.HandleFunc("/xrpc/network.habitat.repo.createRecord", pearServer.CreateRecord)
```
sashankg, line 240: why add this? it's just an additional thing to maintain

cmd/pear/main.go
```diff
@@ -224,30 +232,32 @@ func run(_ context.Context, cmd *cli.Command) error {
...
-	pdsForwarding := newPDSForwarding(pdsCredStore, oauthServer, pdsClientFactory)
-	mux.PathPrefix("/xrpc/").Handler(pdsForwarding)
+	mux.HandleFunc("/xrpc/network.habitat.clique.createClique", cliqueServer.CreateClique)
+	mux.HandleFunc("/xrpc/network.habitat.clique.addMembers", cliqueServer.AddCliqueMembers)
+	mux.HandleFunc("/xrpc/network.habitat.clique.removeMembers", cliqueServer.RemoveCliqueMembers)
+	mux.HandleFunc("/xrpc/network.habitat.clique.getMembers", cliqueServer.GetCliqueMembers)
+	mux.HandleFunc("/xrpc/network.habitat.clique.isMember", cliqueServer.IsCliqueMember)
+
+	pdsForwarding := forwarding.NewPDSForwarding(pdsCredStore, oauthServer, pdsClientFactory, dir)
+	// Only forward specific routes that we know we handle correctly; for now.
```
sashankg, line 257: what was the issue again?

frontend/src/routes/_requireAuth/pear-test/view.tsx
```diff
@@ -17,7 +17,7 @@ export const Route = createFileRoute("/_requireAuth/pear-test/view")({
       params.get("collection") || "network.habitat.test",
     );
     const response = await context.authManager?.fetch(
-      `/xrpc/network.habitat.getRecord?${params.toString()}`,
+      `/xrpc/network.habitat.repo.getRecord?${params.toString()}`,
```
sashankg: if we switch all these to use the `query` function it would be type safe

frontend/src/routes/_requireAuth/_forwarding-test.tsx
```diff
@@ -1,6 +1,6 @@
 import { createFileRoute } from "@tanstack/react-router";
 
-export const Route = createFileRoute("/_requireAuth/forwarding-test")({
+export const Route = createFileRoute("/_requireAuth/_forwarding-test")({
```
sashankg, line 4: why `_forwarding-test` doesn't that hide the path?

typescript/apps/greensky/src/habitatApi.ts
```diff
@@ -1,4 +1,5 @@
-import { Actor, AuthManager, query } from "internal";
+import { AuthManager, getProfile, getProfiles } from "internal";
+export { getProfile, getProfiles };
```
sashankg: why are we exporting them from here?

sashankg (approved):

cmd/pear/main.go
arushibandi: I think it's important to maintain parity if we can:

The remaining com.atproto.repo.* endpoints we don't implement are: applyWrites (batch writes), importRepo (import full public repo as a CAR file -- not sure if / whether we should do this), listMissingBlobs (see here: https://docs.bsky.app/docs/api/com-atproto-repo-apply-writes).

cmd/pear/main.go
arushibandi: We would always forward to the caller's PDS instead of wherever the request is actually supposed to be routed (sometimes target PDS for gets; other times just some other service)

frontend/src/routes/_requireAuth/_forwarding-test.tsx
arushibandi: because it's broken now
before it was looking up the profile, but we don't support that in pds forwarding anymore.

PR #281, opened 2026-04-22
[build]: split up pear vs org entrypoints

arushibandi:
pear is the private data side car + some helpers we're running in production
org is the packaged server orgs can self-host themselves


sashankg (commented):
i don't really understand the point of this?

in what context does the org service exist without pear?

arushibandi:
It's the same thing, we're just doint extra stuff in our prod entrypoint and im adding more under the org deployment i dont want in vanilla pearSent from my iPhoneOn Apr 22, 2026, at 5:59 PM, Sashank Gogula ***@***.***> wrote:﻿
@sashankg commented on this pull request.

i don't really understand the point of this?
in what context does the org service exist without pear?

sashankg:
again there's no "vanilla" pear, right? it's just an org that includes everybody

arushibandi:
i wanna not have to maintain the pear entry point for orgs, but be able to shove random stuff in the main.go, like the waitlist thing. you still need an “if everyone else use this org”, and i’d rather just split up the entrypoints since they are only going to diverge

sashankg:
I can't see how having 2 main.go files that are mostly the same except for orgs is at all the better solution here. I also don't understand what we expect to diverge between pear and org pear. There should be nothing diverging except the org store implementation. 

The waitlist thing is what should be ripped out of pear and put in its own service. Doesn't make sense to copy all the logic and maintain it twice to support use cases like that.

sashankg (approved):
strongly disagree with this approach but approving to unblock since you have stacked changes

PR #282, opened 2026-04-22
[hive]: identity management for orgs

arushibandi:
hive is a service that mints `did:web` identities, serves DID docs / handle mappings, and (in follow-ups), will serve OAuth the way a PDS does. It is so that someone can join a habitat org without having an AT protocol account already, and giving the org control of their account. This is desirable in certain spaces where such an identity only exists in the context of an organization. 

This breaks apart the "you only have one identity" thing that most people have in atproto today. I think there is a way to configure orgs such that people can join with existing identities (and host a data repo within the org for them), but that causes some confusion, like, when you tag someone, are you tagging their "root" or "org" identity? since you can only have one pds. I am bypassing this problem completely by saying orgs mint their own identities. In the future I think there is some way to tie these two together, basically adding a mapping from one identity to another in a verifiable way in someone's PDS, if that's desired. 

This is nice because it gives orgs control of the identity, but still allows these identities to be federated / legible to the outside at proto network. So you could imagine orgs publishing some of their data publicly, and that data needs to have valid `at://` uris etc. And people tagging others within orgs. Doing this achieves that since this is basically implementing the 'auth' part of a PDS. 

The main design choice was directly managing keys / dids instead of using did:plc for a few reasons:
1. did:web: is a bit simpler than did:plc
2. if did:plc goes down, this server is still up which removes did:plc as a SPOF
3. since identities are tied to the organization, no need for it to be in the "public ledger" which is more for public traceability on updates to identities an verifiability things that when you hand over management to an org, don't exist anyways.

Design doc is here, a bit incomplete since i was designing in code / writing / claude https://habitat.network/habitat/docs/#/habitat%3A%2F%2Fdid%3Aplc%3A5czheuspjl4bm56cwash6s7c%2Fnetwork.habitat.docs%2Ff693068e-ba89-4527-b822-8814918d18b8

**Test Plan for this PR:**
- Be able to create an identity (no password, no auth)
- Lookup that identity on `boat.kelinci.net` -- make sure the identity is resolvable and works

Follow-up tests (later PRs):
- Do OAuth with the new identity (always use some test password)
- Add actual password / passkey / ID provider
- TODO: add `@context` field to the DID doc


internal/hive/hive.go
```diff
@@ -0,0 +1,89 @@
...
+
+type hive struct {
+	// TODO: these can't really be changed after the first member is minted; should they be stored in the database for more permanence rather than runtime args?
+
+	// The sub-domain on which members are minted / served for this hive
+	// e.g in "alice.sf.club", this would be "sf.club"
+	domain string
+	// member namespace (subdomain of the domain, or "" if none)
+	// "members" in alice.members.sf.club and "" in alice.sf.club
+	memberNamespace string
```
sashankg: this should just be part of the domain. the server can't publish things under a different namespace besides the domain?

internal/hive/server.go
```diff
@@ -0,0 +1,87 @@
...
+
+func NewServer(hive Hive) (*Server, error) {
+	return &Server{hive: hive}, nil
+}
+
+var (
+	ErrWrongDomain = errors.New("this identity is not hosted by the server")
+)
+
+// returns the identifiery from the host
```
sashankg: typo

internal/hive/server.go
```diff
@@ -0,0 +1,87 @@
...
+type Server struct {
+	hive Hive
+}
+
+func NewServer(hive Hive) (*Server, error) {
+	return &Server{hive: hive}, nil
+}
+
+var (
+	ErrWrongDomain = errors.New("this identity is not hosted by the server")
```
sashankg: it's not really "wrong domain", right? if it was a valid identity it would have to be hosted on this server. so this should just be like "not found"

internal/hive/server.go
```diff
@@ -0,0 +1,87 @@
...
+)
+
+// returns the identifiery from the host
+func (s *Server) parseIdentifier(host string) (syntax.AtIdentifier, error) {
+
+	suffix := "." + s.hive.MemberDomain()
```
sashankg: yeah this seems like a pointless check, right? the server will be served at the host which has to be the member domain

internal/hive/server.go
```diff
@@ -0,0 +1,87 @@
...
+	if !strings.HasSuffix(host, suffix) {
+		return "", ErrWrongDomain
+	}
+
+	return syntax.ParseAtIdentifier(host)
+}
+
+// ServeHTTP routes to ServeDIDDoc or ServeHandle based on the subdomain shape.
```
sashankg: wait don't you serve the diddoc vs handle resolution based on the path? like `.well-known/did.json` you serve the did and `.well-known/atproto-did` you serve the handle

internal/hive/hive.go
arushibandi: it can depending on how you do the DNS aliasing
They could either have a wildcard for `*.sf.club` which i think is what you are understanding OR one alias for `habitat.sf.club` -> habitat server + `*.members.sf.club` -> habitat server so that none of the identities takes up a subdomain.

internal/hive/server.go
arushibandi: Ya this can be not found

arushibandi:
I wasn't able to test this because tailscale won't serve sub-domains. Gonna try to run this on my raspberry pi to test. 

I think we should start figuring out the infra story here sooner rather than later though. Like people should be able to spin up `my-org.habitat.network` and issue identities at `arushi.my-org.habitat.network` for example. or also at `arushi.my.org` and have these alias to `my-org.habitat.network`

cmd/pear/main.go
```diff
@@ -199,10 +205,21 @@ func run(_ context.Context, cmd *cli.Command) error {
 		log.Fatal().Err(err).Msg("unable to setup p2p server")
 	}
 
-	// Order of middlewares = order of "Use" called
-	// https://pkg.go.dev/go.opentelemetry.io/contrib/instrumentation/github.com/gorilla/mux/otelmux
-	mux.Use(otelmux.Middleware("habitat-server"))
-	mux.Use(corsMiddleware)
+	if servingOrg {
```
sashankg: this doesn't need to be gated on `servingOrg`. 

We will always need to have some org specific authz to who can mint identities so the everyone org can just not allow anyone.

the dids and handles will just resolve in not found for ones that are not hosted here (which for the everyone org will be always true)

internal/hive/hive.go
sashankg: yeah then `domain` should be `members.sf.club`, right? why does hive need to now the difference

internal/hive/hive.go
```diff
@@ -0,0 +1,171 @@
...
+
+var _ Hive = &hive{}
+
+// baseDomain returns the full host suffix used in both DIDs and handles:
+// "<memberNamespace>.<domain>" or just "<domain>" if memberNamespace is empty.
+func (h *hive) baseDomain() string {
+	if h.memberNamespace != "" {
+		return h.memberNamespace + "." + h.domain
+	}
+	return h.domain
```
sashankg: yeah we're not using h.domain nor h.memberNamespace anywhere but here. the domain that hive receives should just be the resolved `baseDomain`

internal/hive/hive.go
```diff
@@ -0,0 +1,171 @@
...
+		Handle:      fullHandle,
+		AlsoKnownAs: []string{"at://" + string(fullHandle)},
+		Keys: map[string]identity.VerificationMethod{
+			"atproto": {
+				Type:               "Multikey",
+				PublicKeyMultibase: pub.SigningPublicKey,
+			},
+		},
+		Services: map[string]identity.ServiceEndpoint{
+			"atproto_pds": {
```
sashankg: i can't tell if we should add the atproto_pds or just habitat. fine for now

internal/hive/hive.go
```diff
@@ -0,0 +1,171 @@
...
+				Type: "AtprotoPersonalDataServer",
+				URL:  "https://" + baseDomain,
```
sashankg: hive should take in a separate arg for the pear domain and that's what this should point at. I think that's how to keep this generalizable

internal/hive/hive.go
```diff
@@ -0,0 +1,171 @@
...
+	didStr := did.String()
+	if !strings.HasPrefix(didStr, webPrefix) {
+		return nil, identity.ErrDIDNotFound
+	}
+	host := didStr[len(webPrefix):]
+
+	suffix := "." + h.baseDomain()
+	if !strings.HasSuffix(host, suffix) {
+		return nil, identity.ErrDIDNotFound
+	}
```
sashankg, line 116: imo all this logic should go in the store. the store should take a syntax.DID and return an identity.Identity

internal/hive/hive.go
```diff
@@ -0,0 +1,171 @@
...
+// and looks up the handle against the store.
+func (h *hive) LookupHandle(ctx context.Context, handle syntax.Handle) (*identity.Identity, error) {
+	suffix := "." + h.baseDomain()
+	handleStr := string(handle)
+	if !strings.HasSuffix(handleStr, suffix) {
+		return nil, identity.ErrHandleNotFound
+	}
+
+	internalHandle := handleStr[:len(handleStr)-len(suffix)]
+	pub, err := h.store.getIdentityByHandle(internalHandle)
```
sashankg: same. the store should take the full handle

sashankg (approved):

internal/hive/hive.go
arushibandi: AT proto apps use #atproto_pds to do auth, so we need to do this if people want to sign in to bluesky etc. with this handle. that was going to be my next test after we implement oauth server

internal/hive/hive.go
arushibandi: I was trying to keep the store to not know about `baseDomain` or did doc implementation etc. and just be the backing store.

cmd/pear/main.go
arushibandi: ya i just wasn't doing any auth right now so it was an easy way to protect from someone hitting this API though i guess that's unlikely

internal/hive/hive.go
arushibandi: I did a mix of what you are suggesting and what i had before. I want all the validation etc. to be done in `hive` and then the actual lookup in store. i don't want store to know about domains etc.

PR #284, opened 2026-04-24
[frontend]: ui for inviting org member

arushibandi:
Just testing that this works when hooked up. The user doesn't create a password or anything yet so there is no way for them to be logged in usig this identity but it trivially gets created.

### DIFF
--- a/frontend/src/queries/org.ts
+++ b/frontend/src/queries/org.ts
@@ -64,3 +64,13 @@ export function downgradeAdmin(authManager: AuthManager, admin: string) {
     { authManager },
   );
 }
+
+export function issueInviteToken(authManager: AuthManager) {
+  return procedure(
+    "network.habitat.org.issueInviteToken",
+    {
+      reusable: true,
+    },
+    { authManager },
+  ) as Promise<{ token: string }>;
+}
--- a/frontend/src/queries/org.ts
+++ b/frontend/src/queries/org.ts
@@ -72,5 +72,5 @@ export function issueInviteToken(authManager: AuthManager) {
       reusable: true,
     },
     { authManager },
-  ) as Promise<{ token: string }>;
+  );
 }
--- a/frontend/src/routes/_requireAuth/org/index.tsx
+++ b/frontend/src/routes/_requireAuth/org/index.tsx
@@ -2,14 +2,14 @@ import {
   getAdminsQueryOptions,
   getMembersQueryOptions,
   addAdmin,
-  addMembers,
   removeMembers,
   downgradeAdmin,
+  issueInviteToken,
 } from "@/queries/org";
 import { Button, Input } from "internal";
 import { useMutation, useQueryClient } from "@tanstack/react-query";
 import { createFileRoute, useRouter } from "@tanstack/react-router";
-import { useState } from "react";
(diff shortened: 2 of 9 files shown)


frontend/src/queries/org.ts
```diff
@@ -64,3 +64,13 @@ export function downgradeAdmin(authManager: AuthManager, admin: string) {
...
 }
+
+export function issueInviteToken(authManager: AuthManager) {
+  return procedure(
+    "network.habitat.org.issueInviteToken",
+    {
+      reusable: true,
+    },
+    { authManager },
+  ) as Promise<{ token: string }>;
```
sashankg: never do `as`

frontend/src/routes/_requireAuth/org/index.tsx
```diff
@@ -54,15 +54,70 @@ function OrgPage() {
...
 }
 
+function InviteSection({
+  authManager,
+}: {
+  authManager: Parameters<typeof issueInviteToken>[0];
+}) {
+  const [inviteUrl, setInviteUrl] = useState<string | null>(null);
+  const [copied, setCopied] = useState(false);
+  const inputRef = useRef<HTMLInputElement>(null);
```
sashankg: don't use a ref. use a form

frontend/src/routes/_requireAuth/org/index.tsx
```diff
@@ -54,15 +54,70 @@ function OrgPage() {
...
+        <div className="flex gap-2 mt-3">
+          <Input
+            ref={inputRef}
+            className="flex-1 font-mono text-xs"
+            readOnly
+            value={inviteUrl}
+            onFocus={() => inputRef.current?.select()}
+          />
+          <Button variant="outline" size="sm" onClick={copy}>
+            {copied ? "Copied!" : "Copy"}
```
sashankg: use a sonner for temporary messages like this: https://ui.shadcn.com/docs/components/radix/sonner

frontend/src/routes/org/join.tsx
```diff
@@ -0,0 +1,79 @@
...
+    );
+  }
+
+  return (
+    <div className="flex flex-col gap-4 max-w-md mx-auto mt-16">
+      <h1 className="text-2xl font-semibold">Join Organization</h1>
+      <p className="text-muted-foreground text-sm">
+        Choose a handle for your new account.
+      </p>
+      <form onSubmit={handleSubmit} className="flex flex-col gap-3">
```
sashankg: use react-hook-form. if you're using a form you don't need to track the states yourself

frontend/src/routes/org/join.tsx
```diff
@@ -0,0 +1,79 @@
...
+
+function JoinPage() {
+  const { token } = Route.useSearch();
+  const [handle, setHandle] = useState("");
+  const [result, setResult] = useState<{ handle: string; did: string } | null>(
+    null,
+  );
+  const [error, setError] = useState<string | null>(null);
+  const [loading, setLoading] = useState(false);
+
```
sashankg, line 20: use `useMutation`

sashankg (approved):

PR #288, opened 2026-04-27
[oauth server]: refactor

arushibandi:
Refactor the oauth server to take in login.Router which routes to login.Providers which do the actual flow + deal with any credentials. For now, always allow true from the habitat service login, just so we can start testing this.


sashankg (approved):
Not a big fan of begin and complete. I wonder if there's a more generic interface but can't think of anything off top

PR #295, opened 2026-05-01
[go]: testing style changes

arushibandi:


sashankg (approved):
Is there a linter that can enforce this type of thing?

PR #296, opened 2026-05-01
[repo]: change event emitter for jetstream service

arushibandi:
Eventually this should tail postgres WAL for the repo table, but this is the simplest thing for now. The jetstream service will be the single consumer for the change emitter; coming in the next PR. Everything in the `jetstream` package is unimplemented, just defining the rough outline to know how things look like.


internal/jetstream/server.go
```diff
@@ -0,0 +1,83 @@
...
+func (s *Server) HandleSubscribe(w http.ResponseWriter, r *http.Request) {
+	// Set required SSE headers
+	w.Header().Set("Content-Type", "text/event-stream")
+	w.Header().Set("Cache-Control", "no-cache")
+	w.Header().Set("Connection", "keep-alive")
+
+	// Ensure the response writer supports flushing
+	flusher, ok := w.(http.Flusher)
+	if !ok {
+		http.Error(w, "Streaming unsupported!", http.StatusInternalServerError)
```
sashankg: Use the error util to also log?

internal/repo/changelog.go
```diff
@@ -0,0 +1,86 @@
...
+	"sync/atomic"
+	"time"
+
+	"github.com/bluesky-social/jetstream/pkg/models"
+	"github.com/bradenaw/juniper/stream"
+	"github.com/rs/zerolog/log"
+)
+
+type changeEmitter struct {
+	ctx context.Context
```
sashankg, line 16: The methods should take in a context not the cosntructor

internal/repo/repo.go
```diff
@@ -92,11 +100,13 @@ type link struct {
 }
 
 // TODO: create table etc.
-func NewRepo(db *gorm.DB) (*repo, error) {
+func NewRepo(ctx context.Context, db *gorm.DB) (*repo, error) {
```
sashankg: Yeah NewRepo should not take a context. Just the methods

sashankg (approved):

internal/repo/changelog.go
arushibandi: No in this case the constructor should because the DB request context is tied to the http context, but this one should be more long lived.

PR #299, opened 2026-05-07
[orgs]: require passwords for new identities

arushibandi:


arushibandi:
https://github.com/user-attachments/assets/5662ec08-0e90-41f4-a3fc-810e1e6559e5

cmd/pear/main.go
```diff
@@ -209,7 +234,7 @@ func run(_ context.Context, cmd *cli.Command) error {
 	mux.HandleFunc("/xrpc/network.habitat.org.getAdmins", orgServer.GetAdmins)
 	mux.HandleFunc("/xrpc/network.habitat.org.getMembers", orgServer.GetMembers)
 	mux.HandleFunc("/xrpc/network.habitat.org.addAdmin", orgServer.AddAdmin)
-	mux.HandleFunc("/xrpc/network.habitat.org.addMembers", orgServer.AddMembers)
+	// mux.HandleFunc("/xrpc/network.habitat.org.addMembers", orgServer.AddMembers)
```
sashankg: why comment this out?

frontend/src/routes/login/habitat.tsx
```diff
@@ -1,11 +1,69 @@
...
+  const {
+    register,
+    handleSubmit,
+    setError,
+    formState: { isSubmitting, errors },
+  } = useForm<FormValues>();
+
+  const onSubmit = async ({ password }: FormValues) => {
+    try {
+      const res = await fetch(`https://${__HABITAT_DOMAIN__}/xrpc/network.habitat.org.loginMember`, {
```
sashankg: use `procedure`

frontend/src/routes/login/habitat.tsx
```diff
@@ -1,11 +1,69 @@
...
+          <Input
+            type="password"
+            placeholder="password"
+            disabled={isSubmitting}
+            {...register("password", { required: true })}
+          />
+          <FieldError errors={[errors.password]} />
+        </Field>
+        <FieldError errors={[errors.root]} />
+        <Button type="submit" disabled={isSubmitting}>
```
sashankg: pretty sure you can just disable the entire form

frontend/src/routes/org/join.tsx
```diff
@@ -25,13 +25,12 @@ function JoinPage() {
     formState: { isSubmitting, errors },
   } = useForm<FormValues>();
 
-  const onSubmit = async ({ handle }: FormValues) => {
+  const onSubmit = async ({ handle, password }: FormValues) => {
     try {
-      // TODO: is this the right way to target habitat domain ?
       const res = await fetch(`https://${__HABITAT_DOMAIN__}/xrpc/network.habitat.org.mintMemberIdentity`, {
```
sashankg: use `procedure` here too

frontend/src/routes/__root.tsx
```diff
@@ -22,22 +22,32 @@ export const Route = createRootRouteWithContext<RouterContext>()({
...
-    const [config, profileResponse] = await Promise.all([
+    const [config, profileResult] = await Promise.allSettled([
       context.queryClient.fetchQuery(getConfigQueryOptions(context.authManager)),
       new AtpAgent({ service: "https://public.api.bsky.app" }).getProfile({
         actor: authInfo.did,
       }),
     ]);
 
-    return { profile: profileResponse.data, org: config };
+    const profile =
```
sashankg, line 32: why are we doing all this? if the profile fetch fails we should just throw an error, right? If there's some condition we expect it to fail then we should just not make the request. don't let it fail and try to handle the failure

frontend/src/routes/__root.tsx
```diff
@@ -22,22 +22,32 @@ export const Route = createRootRouteWithContext<RouterContext>()({
...
+      profile,
+      org: config.status === "fulfilled" ? config.value : undefined,
+    };
   },
   staleTime: 1000 * 60 * 60,
   component() {
     const { authManager } = Route.useRouteContext();
     const { profile, org } = Route.useLoaderData();
+    const pathname = useRouterState({ select: (s) => s.location.pathname });
+    const showHeader = !pathname.startsWith("/login");
```
sashankg: we shouldn't do stuff like this. if we want to conditionally render the header then we put it in an intermediate route that wraps the routes that we do want headers for. 

do we not want to show the header because they're not signed in? then the header should only be in the `_requireAuth` route. 

otherwise just leave it on the login page for now

internal/hive/hive.go
```diff
@@ -20,8 +20,8 @@ var handlePattern = regexp.MustCompile(`^[a-zA-Z0-9]{1,50}$`)
 
 type Hive interface {
 	// Minting new identities for members
-	MintIdentity(handle string) (*identity.Identity, error)
-
+	MintIdentity(handle string) (*identity.Identity, func(*gorm.DB) error, error)
+	BaseDomain() string
```
sashankg: this feels like a leaky abstraction especially since it's only used by wrapped directory. Maybe hive itself can implement directory or it can have a method that takes the default directory and returns a wrapped directory.

internal/hive/wrapped_dir.go
```diff
@@ -0,0 +1,93 @@
...
+		return err
+	}
+	return dir.Purge(ctx, atid)
+}
+
+var _ identity.Directory = &wrappedDir{}
+
+func NewWrappedDirectory(hive Hive, fallback identity.Directory) identity.Directory {
+	return &wrappedDir{
+		fallback: fallback,
```
sashankg, line 52: do we really need the fallback?

internal/oauthserver/oauth_server.go
```diff
@@ -182,7 +181,7 @@ type OAuthServer struct {
 //
 // Returns a configured OAuthServer ready to handle authorization requests.
 func NewOAuthServer(
-	secret string,
+	secret []byte,
 	loginRouter *login.Router,
```
sashankg, line 185: i'm realizing now that we should have a separate interface for login methods. the pds, loginProvider, and eventually SSO providers will implement that interface and the oauth server just invokes one of them based on the identity

sashankg (approved):

cmd/pear/main.go
arushibandi: you can only add members by creating a new member identity now

frontend/src/routes/__root.tsx
arushibandi: there is no profile for org identities. since this fetches `app.bsky.profile`

we need to handle this better but for now i wanted to keep the non-org profile photo working and not fail when this wasn't existing for org members

frontend/src/routes/__root.tsx
arushibandi: ok i can remove it for now, eventually this page needs to be nicer

internal/hive/wrapped_dir.go
arushibandi: i think so. for e.g. to resolve requesters not in the org etc.

internal/hive/hive.go
arushibandi: hive does implement directory

internal/oauthserver/oauth_server.go
arushibandi: ya that's what i was trying to do with `login.Provider` -- make that the interface that all the logins implement

arushibandi: and the `login.Router` picks which one to use

internal/hive/hive.go
arushibandi: made this a bit less leaky

PR #302, opened 2026-05-08
[orgs]: fixes

arushibandi:
I messed up the login page when i replaced to use procedure and also i realized we should use opaqueIDs and not dids in the members table because the web domain part could change.

### DIFF
--- a/frontend/src/routes/login/habitat.tsx
+++ b/frontend/src/routes/login/habitat.tsx
@@ -27,7 +27,7 @@ function HabitatLoginPage() {
       const { callbackURL } = await procedure(
         "network.habitat.org.loginMember",
         { handle, password },
-        { domain: __HABITAT_DOMAIN__ },
+        { unauthenticated: true, domain: __HABITAT_DOMAIN__ },
       );
       window.location.href = `https://${__HABITAT_DOMAIN__}${callbackURL}`;
     } catch (err) {
--- a/frontend/src/routes/login/habitat.tsx
+++ b/frontend/src/routes/login/habitat.tsx
@@ -1,7 +1,6 @@
 import { Button, Field, FieldError, FieldLabel, Input } from "internal/components/ui";
 import { createFileRoute } from "@tanstack/react-router";
 import { useForm } from "react-hook-form";
-import { procedure } from "internal";
 
 export const Route = createFileRoute("/login/habitat")({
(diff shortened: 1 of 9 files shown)


frontend/src/routes/login/habitat.tsx
```diff
@@ -25,11 +23,22 @@ function HabitatLoginPage() {
 
   const onSubmit = async ({ password }: FormValues) => {
     try {
-      const { callbackURL } = await procedure(
-        "network.habitat.org.loginMember",
-        { handle, password },
-        { authManager },
```
sashankg, line 31: Let's just add an unauthenticated param to skip all the auth logic

sashankg: Bonus points if we can add it to the endpoint object so procedure always know loginMember doesn't require auth

internal/hive/hive.go
```diff
@@ -20,7 +20,7 @@ var handlePattern = regexp.MustCompile(`^[a-zA-Z0-9]{1,50}$`)
 
 type Hive interface {
 	// Minting new identities for members
-	MintIdentity(handle string) (*identity.Identity, func(*gorm.DB) error, error)
+	MintIdentity(handle string) (*identity.Identity, string /* opaqueID */, func(*gorm.DB) error, error)
```
sashankg: Hmm if we can't rely on the did remaining stable I think that is going to be annoying in a lot of places. Fine for now I guess but this seems like a can of worms

sashankg (approved):

internal/hive/hive.go
arushibandi: I agree i also don't like it
Maybe that's why roundabout ran their own plc directory; we could do something similar and possibly more lightweight. idk.

This only becomes an issue if an org moves servers and doesn't want the `.habitat.network` domain tied to their org specifically on DIDs (like they could have handles with a different domain map to the `.habitat.network` did. So not sure if it's worth pursuing rn. But if an org wants to go from managed hosting --> self-hosting it would be an issue,

arushibandi: @sashankg 
lmk what you think about this, at some point. gonna merge for now

arushibandi: or i can hold off, i feel like maybe this is wrong

PR #306, opened 2026-05-11
[org passwords]: fix routing

arushibandi:
I think it should work after this


sashankg (approved):
why are we using the handle? that's definitely wrong, right? approving to unblock

arushibandi:
> why are we using the handle? that's definitely wrong, right? approving to unblock

handle is `arushi.test-org.habitat.network`

PR #307, opened 2026-05-11
[Orgs] Store for multitenancy

sashankg:
doesn't actually let you create stores yet but reorganizes the read path to support multiple stores


internal/org/org.go
```diff
@@ -31,21 +30,13 @@ var (
...
 	ErrInvalidTokenExpiry = errors.New("token expiry must be < 1 month from now")
+	ErrOrgNotFound        = errors.New("organization not found")
+	ErrMemberNotFound     = errors.New("member not found in any org")
 )
 
-// Store is used to manage the organization tied to a pear node
-// It enforces permissions on its methods; i.e non-admins cannot add admins or members.
-//
-// Eventually we can add management scopes to member vs. admin roles, to allow members to add other members etc.
+// Org represents a single organization on a pear instance.
```
arushibandi, line 38: is it you deleting comments or claude? claude always deletes my comments including things i actually need

internal/org/org.go
```diff
@@ -54,43 +45,42 @@ type Org interface {
...
 	CreateNewMemberIdentity(ctx context.Context, token string, internalHandle string, password string) (*identity.Identity, error)
 	AuthenticateMember(ctx context.Context, handle string, password string) (bool, error)
 
-	// Generic config about this org
 	GetMetadata() habitat.NetworkHabitatOrgGetMetadataOutput
 }
 
+// Store is the registry of all orgs on a pear instance.
+// It routes DIDs to their org and provides cross-org membership checks.
+type Store interface {
```
arushibandi, line 77: can we call it `Registry`? since we use `store` to mean the db backing store in other places

internal/org/org.go
```diff
@@ -54,43 +45,42 @@ type Org interface {
...
+	GetOrg(ctx context.Context, orgID string) (Org, error)
+	GetOrgByDID(ctx context.Context, did syntax.DID) (Org, error)
```
arushibandi: nit: `GetOrgForDID` right? since it's a user DID not an org did?

internal/org/org.go
```diff
@@ -54,43 +45,42 @@ type Org interface {
...
+	AuthenticateMember(ctx context.Context, handle string, password string) (bool, error)
```
arushibandi, line 80: does this need to take in orgID now?

internal/org/org.go
```diff
@@ -54,43 +45,42 @@ type Org interface {
...
+}
+
 type inviteTokenClaims struct {
 	jwt.Claims
 	Reusable bool `json:"reusable"`
 }
 
 type store struct {
```
arushibandi: we could name this `org` to make it less confusing

internal/oauthserver/oauth_server.go
```diff
@@ -382,14 +382,9 @@ func (o *OAuthServer) HandleCallback(
 
 	// Check the DID allowlist after reconstructing authRequest so we can redirect
 	// errors back to the client via WriteAuthorizeError instead of returning a raw 401.
-	allowed, err := o.org.IsMember(r.Context(), arf.Did)
+	_, err = o.orgStore.GetOrgByDID(r.Context(), arf.Did)
```
arushibandi: why don't we use the return value?

internal/oauthserver/oauth_server.go
```diff
@@ -499,11 +494,8 @@ func (o *OAuthServer) Validate(
 		return "", false
 	}
 
-	ok, err = o.org.IsMember(r.Context(), did)
+	_, err = o.orgStore.GetOrgByDID(r.Context(), did)
 	if err != nil {
```
arushibandi, line 498: why did we delete a path?

internal/org/models.go
```diff
@@ -2,8 +2,18 @@ package org
 
 import "time"
 
-// Keep track of members in the org
+// Organization represents a managed org on a pear instance.
+type Organization struct {
```
arushibandi, line 6: i don't think this needs to be exported?

internal/org/org.go
```diff
@@ -312,26 +302,102 @@ func (s *store) CreateNewMemberIdentity(ctx context.Context, token string, inter
 
 func (s *store) AuthenticateMember(ctx context.Context, handle string, password string) (bool, error) {
 	id, err := s.hive.LookupHandle(ctx, syntax.Handle(handle))
-	if errors.Is(err, identity.ErrHandleNotFound) || errors.Is(err, identity.ErrInvalidHandle) {
-		// Don't leak whether the handle exists
+	if err != nil {
 		return false, nil
-	} else if err != nil {
-		return false, err
 	}
```
arushibandi, line 383: shouldn't there be new auth changes to `CreateNewMemberIdentity` now? since you made the API change to add orgID as a url param

arushibandi (approved):
i think claude is deleting a lot of comments that i probably want to keep

internal/oauthserver/oauth_server.go
sashankg: we confirm that the did belongs to an org by looking up the org by the did. it'll error if there's no org for that did. we don't need to additionally check if they are a member of the org

internal/org/org.go
sashankg: this was ai. will restore the ones that were useful

internal/oauthserver/oauth_server.go
sashankg: don't need the actual org. just need to check if the did is in a org

internal/org/models.go
sashankg: i'm using it in a test file. will fix when we add proper org creation in follow up

internal/org/org.go
sashankg: "store" is the name of the interfaces and the db is part of the implementation. I don't think we should saying all "stores" are tied to a specific implementation. "Store" interfaces are anything that has getters and setters

internal/org/org.go
sashankg: yeah will fix in follow up

PR #309, opened 2026-05-13
[Org] Create

sashankg:


internal/org/org.go
```diff
@@ -78,6 +79,12 @@ type Store interface {
 	GetOrg(ctx context.Context, orgID string) (Org, error)
 	GetOrgForDID(ctx context.Context, did syntax.DID) (Org, error)
 	AuthenticateMember(ctx context.Context, handle string, password string) (bool, error)
+	CreateOrg(
+		ctx context.Context,
+		name string,
+		adminHandle string,
+		adminPassword string,
+	) (orgID string, id *identity.Identity, err error)
```
arushibandi, line 88: We should gate this at first so only allowed test users can sign up.

api/habitat/org_create.go
```diff
@@ -0,0 +1,18 @@
+package habitat
+
+// Code generated by lexgen. DO NOT EDIT.
+
+// NetworkHabitatOrgCreateInput represents the input for network.habitat.org.create
+type NetworkHabitatOrgCreateInput struct {
+	AdminHandle   string `json:"admin_handle"`
+	AdminPassword string `json:"admin_password"`
+	Name          string `json:"name,omitempty"`
```
arushibandi, line 9: can this be empty though?

cmd/pear/main.go
```diff
@@ -217,19 +217,26 @@ func run(_ context.Context, cmd *cli.Command) error {
...
 	mux.HandleFunc("/xrpc/network.habitat.org.getAdmins", orgServer.GetAdmins)
 	mux.HandleFunc("/xrpc/network.habitat.org.getMembers", orgServer.GetMembers)
 	mux.HandleFunc("/xrpc/network.habitat.org.addAdmin", orgServer.AddAdmin)
 	mux.HandleFunc("/xrpc/network.habitat.org.removeAdmin", orgServer.RemoveAdmin)
 	mux.HandleFunc("/xrpc/network.habitat.org.removeMembers", orgServer.RemoveMembers)
 	mux.HandleFunc("/xrpc/network.habitat.org.downgradeAdmin", orgServer.DowngradeAdmin)
+	mux.HandleFunc("/xrpc/network.habitat.org.create", orgServer.CreateOrg)
 	mux.HandleFunc("/xrpc/network.habitat.org.issueInviteToken", orgServer.IssueInviteToken)
 	mux.HandleFunc("/xrpc/network.habitat.org.mintMemberIdentity", orgServer.MintMemberIdentity)
+	mux.HandleFunc("/xrpc/network.habitat.org.create", orgServer.CreateOrg)
```
arushibandi, line 228: why is it duplicated

internal/org/server.go
```diff
@@ -43,6 +44,48 @@ func (s *Server) BootstrapAdmin(w http.ResponseWriter, r *http.Request) {
...
 
+func (s *Server) CreateOrg(w http.ResponseWriter, r *http.Request) {
+	// no auth: bootstrapping a new org
+	var req habitat.NetworkHabitatOrgCreateInput
+	if err := json.NewDecoder(r.Body).Decode(&req); err != nil {
+		utils.LogAndHTTPError(w, err, "reading request body", http.StatusBadRequest)
+		return
+	}
+
+	if req.AdminHandle == "" || req.AdminPassword == "" || req.Name == "" {
```
arushibandi, line 55: ya name shouldn't have `omitEmpty` then right? bc its not optional?

api/habitat/org_mintMemberIdentity.go
```diff
@@ -5,7 +5,7 @@ package habitat
 // NetworkHabitatOrgMintMemberIdentityInput represents the input for network.habitat.org.mintMemberIdentity
 type NetworkHabitatOrgMintMemberIdentityInput struct {
 	Handle   string `json:"handle"`
-	OrgId    string `json:"org_id"`
+	OrgId    string `json:"orgId,omitempty"`
```
arushibandi, line 8: why is this omitempty too?
i guess it doesn't matter if its checked serversisde just confused about this change

arushibandi (approved):

internal/org/org.go
sashankg: we can add it in a follow up. i don't think we should be concerned with randos signing up now. we can just wipe the db

api/habitat/org_create.go
sashankg: fixed

cmd/pear/main.go
sashankg: oops

api/habitat/org_mintMemberIdentity.go
sashankg: the org id i think is temporary for now. I'm still figuring out the right way to do this. will revisit

PR #310, opened 2026-05-13
[Handles] New service

sashankg:
new service to decouple handles from pear. an org can just serve its dids from pear under `did:web:*.<pear-domain>`. handles which are more useful to be aesthetic can be served somewhere else and just need to point to the right did


internal/handles/store.go
```diff
@@ -0,0 +1,53 @@
...
+)
+
+type handleMapping struct {
+	Handle string `gorm:"primaryKey"`
+	DID    string
+}
+
+type store struct {
+	db *gorm.DB
+}
```
arushibandi, line 19: why have this nested handlesImpl -> store -> DID ? can we just put db *gorm.DB directly in handlesImpl?

internal/handles/server.go
```diff
@@ -0,0 +1,69 @@
...
+	_, _ = w.Write([]byte(ident.DID.String()))
+}
+
+func (s *Server) MintHandle(w http.ResponseWriter, r *http.Request) {
+	if r.Method != http.MethodPost {
+		http.Error(w, "method not allowed", http.StatusMethodNotAllowed)
+		return
+	}
+
+	// TODO add auth
```
arushibandi, line 44: this will be the token auth right?

arushibandi (approved):

internal/handles/server.go
sashankg: yeah still figuring that out

PR #315, opened 2026-05-14
[Typescript] Fix formatting

sashankg:


.moon/tasks/typescript.yml
```diff
@@ -8,7 +8,6 @@ tasks:
       - '@in(0)'
       - --ignore-path
       - '@in(1)'
-      - --write
```
sashankg, line 11: it was writing during CI so CI wasn't complaining about fixable issues

arushibandi (approved):

PR #318, opened 2026-05-14
[AuthZen] Evaluate API

sashankg:
- **feat: add authzen.evaluate lexicon and generated types**
- **feat: add authzen evaluate server with handler and tests**
- **fix: address code review - log unsupported actions, fix body close in test**
- **feat: wire authzen evaluate endpoint into pear server**

https://openid.net/specs/authorization-api-1_0.html


lexicons/network/habitat/authzen/evaluate.json
```diff
@@ -0,0 +1,91 @@
...
+      "properties": {
+        "type": {
+          "type": "string"
+        },
+        "id": {
+          "type": "string"
+        },
+        "properties": {
+          "type": "object"
+        }
```
arushibandi: should we move this into a #def bc it's the same one as below

arushibandi: also in the server you only check `subject.id` and `object.id` not the `type` property so is it required?

lexicons/network/habitat/authzen/evaluate.json
```diff
@@ -0,0 +1,91 @@
...
+      "type": "object",
+      "required": ["name"],
+      "properties": {
+        "name": {
+          "type": "string"
+        },
+        "properties": {
+          "type": "object"
+        }
+      }
```
arushibandi, line 35: why not just make this a string? what is nested properties

lexicons/network/habitat/authzen/evaluate.json
```diff
@@ -0,0 +1,91 @@
...
+    },
+    "action": {
+      "type": "object",
+      "required": ["name"],
+      "properties": {
+        "name": {
+          "type": "string"
+        },
+        "properties": {
+          "type": "object"
```
arushibandi: can you add descriptions / formats here where possible

arushibandi (approved):

lexicons/network/habitat/authzen/evaluate.json
sashankg: Addressed this by creating a shared `entity` def in `defs` and making both `subject` and `resource` in `main` ref `#entity`. (Note: AT Protocol lexicon spec prohibits ref-type top-level defs, so `subject`/`resource` can't be ref-only defs — instead `main` references `#entity` directly for both.)

lexicons/network/habitat/authzen/evaluate.json
sashankg: Per the AuthZEN spec (Section 5.3), `properties` on an Action is explicitly defined as an `object`: "an OPTIONAL properties key with an object value." The spec example uses it for things like `{"method": "GET"}`.

lexicons/network/habitat/authzen/evaluate.json
sashankg: Done — added descriptions per the AuthZEN spec to all fields in the lexicon.

lexicons/network/habitat/authzen/evaluate.json
sashankg: @arushibandi — re: whether `type` should be required in the entity def when the server only validates `id`: this is an API interface decision. The AuthZEN spec requires both `type` and `id` as REQUIRED, but open to discussion.

sashankg:
sorry i let AI handle the PR comments. I'm not gonna merge this yet since it's lower pri. will revist

PR #319, opened 2026-05-15
[FGA] Store

sashankg:
Basic wrapper around OpenFGA that will be used to resolve permission queries with spaces. For now, the authorization model just has "organization", "user" and "space". That should be enough to get simple permissions working. we will update the model to support user groups and space groups such that they can create inheritance relationships


docs/superpowers/specs/2026-05-14-fgastore-design.md
```diff
@@ -0,0 +1,228 @@
+# fgastore: OpenFGA-backed Authorization Store
+
+**Goal:** Define an authorization model (user, record, org, role as object types) and expose a `Store` interface wrapping OpenFGA's storage layer, usable by the AuthZEN evaluator.
```
arushibandi, line 3: in the design doc, we should explicitly call out our constraints / what use cases we need to support and now support. e.g for application-defined roles, some example use-cases motivated by calendar / docs / code management apps just so we can cover our bases. 

also, should we look into teams / organizational hierarchies? are these roles? i need to walk through examples to understand it fully

docs/superpowers/specs/2026-05-14-fgastore-design.md
```diff
@@ -0,0 +1,228 @@
...
+---
+
+## Types & Relations
+
+Four types with direct-only relations (`[user]` type restrictions — no usersets or computed relations yet):
+
+| Type | Relations | Description |
+|------|-----------|-------------|
+| `user` | _(none)_ | A DID (e.g. `user:did:plc:abc123`) |
+| `record` | `owner`, `can_read` | An AT URI record |
```
arushibandi, line 18: i'm a little confused about this ... ideally we are not storing any data in two places that need to be kept in sync -- i.e. we probably shouldn't store record can_read somewhere and a permission space member list elsewhere. asme thing for record-owner; isn't this encoded naturally in the rows of the repo table?

docs/superpowers/specs/2026-05-14-fgastore-design.md
```diff
@@ -0,0 +1,228 @@
+# fgastore: OpenFGA-backed Authorization Store
+
+**Goal:** Define an authorization model (user, record, org, role as object types) and expose a `Store` interface wrapping OpenFGA's storage layer, usable by the AuthZEN evaluator.
+
+**Architecture:** A new `internal/fgastore/` package wraps OpenFGA's `storage.OpenFGADatastore` behind a habitat-specific interface. The constructor picks the backend (PostgreSQL for production, in-memory for tests). The auth model is defined as OpenFGA `TypeDefinition` structs at startup. Check/Write/Delete delegate to the storage layer's direct tuple operations.
```
arushibandi, line 5: should we define the auth model in lexicons / records ? is this the relationship tuple thing you were talking about ?

docs/superpowers/specs/2026-05-14-fgastore-design.md
```diff
@@ -0,0 +1,228 @@
...
+    Write(ctx context.Context, user, relation, object string) error
+    Delete(ctx context.Context, user, relation, object string) error
+    Close() error
+}
+```
+
+All parameters are plain strings. Callers construct values like:
+- `user`: `"user:did:plc:abc123"` or `"user:*"` (future: `"org:uuid#member"`)
+- `relation`: `"can_read"`, `"owner"`, `"member"`, `"admin"`, `"assigned"`
+- `object`: `"record:did:plc:abc123/net.habitat.post/3kq..."`, `"org:uuid"`, `"role:admin"`
```
arushibandi, line 45: would be nice to make these `syntax.DID` and named strings for the purpose of making the API more readable

internal/fgastore/fgastore.go
```diff
@@ -0,0 +1,152 @@
...
+	ds, err := postgres.New(uri, &sqlcommon.Config{})
+	if err != nil {
+		return nil, fmt.Errorf("fgastore postgres: %w", err)
+	}
+	return newFromDS(ctx, ds)
+}
+
+func NewInMemory(ctx context.Context) (*FGA, error) {
+	return newFromDS(ctx, memory.New())
+}
```
arushibandi, line 44: wait why can't you give it a `gorm.DB` or always give it `storage.OpenFGADatastore`? this seems wrong

internal/fgastore/fgastore.go
```diff
@@ -0,0 +1,152 @@
...
+	Delete(ctx context.Context, user, relation, object string) error
+	ListObjects(ctx context.Context, user, relation, objectType string) ([]string, error)
+	ListUsers(ctx context.Context, object, relation string) ([]string, error)
+	Close() error
+}
+
+type FGA struct {
+	ds      storage.OpenFGADatastore
+	svr     *server.Server
+	storeID string
```
arushibandi, line 31: do we need a store per-org ?

internal/fgastore/fgastore.go
```diff
@@ -0,0 +1,152 @@
...
+)
+
+var _ Store = (*FGA)(nil)
+
+type Store interface {
+	Check(ctx context.Context, user, relation, object string) (bool, error)
+	Write(ctx context.Context, user, relation, object string) error
+	Delete(ctx context.Context, user, relation, object string) error
+	ListObjects(ctx context.Context, user, relation, objectType string) ([]string, error)
+	ListUsers(ctx context.Context, object, relation string) ([]string, error)
```
arushibandi, line 24: ya can we formaloize these types a bit instead of just strings?

internal/fgastore/model.go
```diff
@@ -0,0 +1,182 @@
+package fgastore
+
+import (
+	openfgav1 "github.com/openfga/api/proto/openfga/v1"
+)
+
+func authModel() *openfgav1.AuthorizationModel {
+	return &openfgav1.AuthorizationModel{
+		SchemaVersion: "1.1",
+		TypeDefinitions: []*openfgav1.TypeDefinition{
```
arushibandi, line 10: wonder if we should do the `owner`/`admin`/`member` thing now (which i believe solves the issue of malicious admin deleting all other admins or something?)

internal/fgastore/model.go
```diff
@@ -0,0 +1,182 @@
...
+							},
+						},
+					},
+				},
+			},
+			{
+				Type: "space",
+				Relations: map[string]*openfgav1.Userset{
+					"owner":  {Userset: &openfgav1.Userset_This{}},
+					"member": {Userset: &openfgav1.Userset_This{}},
```
arushibandi, line 50: does it need to be unioned with the owner?

internal/fgastore/model.go
```diff
@@ -0,0 +1,182 @@
...
+						},
+					},
+					"can_manage_members": {
+						Userset: &openfgav1.Userset_Union{
+							Union: &openfgav1.Usersets{Child: []*openfgav1.Userset{
+								{Userset: &openfgav1.Userset_This{}},
+								{
+									Userset: &openfgav1.Userset_ComputedUserset{
+										ComputedUserset: &openfgav1.ObjectRelation{
+											Relation: "owner",
```
arushibandi, line 114: an editor can't manage the space ? 
what about case where you want people who you shared doc with to be able to share with other people ? do we need scopes here or something ?

internal/fgastore/model.go
```diff
@@ -0,0 +1,182 @@
...
+										},
+									},
+								},
+							}},
+						},
+					},
+				},
+				Metadata: &openfgav1.Metadata{
+					Relations: map[string]*openfgav1.RelationMetadata{
+						"owner": {
```
arushibandi, line 124: can u explain this to me

docs/superpowers/specs/2026-05-14-fgastore-design.md
sashankg: yeah the authorization model just supports those relationships. it doesn't mean we need to store anything additional. when it comes time to query permissions, we provide the context from the source of truth

docs/superpowers/specs/2026-05-14-fgastore-design.md
sashankg: the auth model here is an fga store implementation detail. we will additionally expose an api for callers to contribute to the store which will look like openfga tuples but we can enforce validation on them

docs/superpowers/specs/2026-05-14-fgastore-design.md
sashankg: these are all implementation details to the fga store. won't be exposed to callers

internal/fgastore/fgastore.go
sashankg: oh yeah it should take the the db as input my bad.

internal/fgastore/fgastore.go
sashankg: i don't think so. we just need to validate before adding relationships that people aren't adding permissions across orgs.

internal/fgastore/model.go
sashankg: will keep it simple for now. we can change this whenever

internal/fgastore/fgastore.go
sashankg: actually nvm. openfga doesn't just take in a sql.DB. we'll have to pass in the url and create a new connection pool but i think that's fine.

internal/fgastore/fgastore.go
sashankg: will follow up on this later. the "user" here isn't necessarily a did. it could be "org#member". Will figure out the right api when we have more built out

internal/fgastore/model.go
sashankg: not over indexing on the auth model rn. just want to get something out and we can iterate on it later

arushibandi (approved):

PR #320, opened 2026-05-15
[Auth] Support auth method per org

sashankg:
- **feat: add LoginMethod field to org model and interface**
- **test: add LoginMethod coverage for everyoneOrg, orgImpl, and CreateOrg**
- **refactor: replace Type/CanHandle on LoginProvider with LoginMethod**
- **refactor: replace ProviderType/CanHandle with LoginMethod-based routing**
- **feat: wire mapping store into pdsProvider Authorize with test**
- **feat: add googleProvider stub**
- **refactor: use pdsclient.DummyDirectory in mapping test, remove unused test helpers**
- **refactor: route OAuth providers by org login method, wire main.go**
- **commit plan**

-------------------------
- master
  - **[Auth] Support auth method per org** :point_left:
    - #326
    - #325

### DIFF
--- a/internal/login/pds.go
+++ b/internal/login/pds.go
@@ -15,9 +15,15 @@ import (
 	"github.com/rs/zerolog/log"
 )
 
+type mappingStore interface {
+	GetPublicDID(ctx context.Context, did syntax.DID) (*syntax.DID, error)
+}
+
 type pdsProvider struct {
-	oauthClient pdsclient.PdsOAuthClient
-	credStore   pdscred.PDSCredentialStore
+	oauthClient  pdsclient.PdsOAuthClient
+	credStore    pdscred.PDSCredentialStore
+	mappingStore mappingStore
+	dir          identity.Directory
 }
 
 // pdsProviderState is the opaque flash state for the PDS login flow.
@@ -26,17 +32,11 @@ type pdsProviderState struct {
 	AuthorizeState pdsclient.AuthorizeState `json:"authorize_state"`
 }
 
-func NewPDSProvider(oauthClient pdsclient.PdsOAuthClient, credStore pdscred.PDSCredentialStore) Provider {
-	return &pdsProvider{oauthClient: oauthClient, credStore: credStore}
+func NewPDSProvider(oauthClient pdsclient.PdsOAuthClient, credStore pdscred.PDSCredentialStore, ms mappingStore, dir identity.Directory) Provider {
+	return &pdsProvider{oauthClient: oauthClient, credStore: credStore, mappingStore: ms, dir: dir}
 }
 
-func (p *pdsProvider) Type() ProviderType { return ProviderTypePDS }
-
-func (p *pdsProvider) CanHandle(id *identity.Identity) bool {
-	_, hasPDS := id.Services["atproto_pds"]
-	_, hasHabitat := id.Services["habitat"]
-	return hasPDS && !hasHabitat
-}
+func (p *pdsProvider) LoginMethod() string { return "atproto" }
 
 func (p *pdsProvider) Authorize(_ context.Context, id *identity.Identity) (string, []byte, error) {
 	dpopKey, err := ecdsa.GenerateKey(elliptic.P256(), rand.Reader)
--- a/internal/login/pds.go
+++ b/internal/login/pds.go
@@ -10,6 +10,7 @@ import (
 
 	"github.com/bluesky-social/indigo/atproto/identity"
 	"github.com/bluesky-social/indigo/atproto/syntax"
+	"github.com/habitat-network/habitat/internal/org"
 	"github.com/habitat-network/habitat/internal/pdsclient"
 	"github.com/habitat-network/habitat/internal/pdscred"
 	"github.com/rs/zerolog/log"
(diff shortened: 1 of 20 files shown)


internal/org/models.go
```diff
@@ -6,17 +6,18 @@ import "time"
 type organization struct {
 	ID            string `gorm:"primaryKey"`
 	Name          string // optional display name
+	LoginMethod   string // "atproto", "google", "password"
```
arushibandi: make
```
type LoginMethod string

const (
    LoginMethodPassword = "password"
    LoginMethodATProtocol = "aproto"
    ...
)

type organization struct {
	ID                     string `gorm:"primaryKey"`
	Name               string // optional display name
	LoginMethod   LoginMethod // "atproto", "google", "password"
    ...
}
```

internal/org/models.go
```diff
@@ -6,17 +6,18 @@ import "time"
...
-	OrgID        string `gorm:"primaryKey"`
-	Member       string `gorm:"primaryKey"`
-	Role         string
-	PasswordHash string `gorm:"not null"`
+// Each Member belongs to exactly one org.
+type Member struct {
+	OrgID   string `gorm:"primaryKey"`
+	Member  string `gorm:"primaryKey"`
+	Role    string
+	LoginID string // provider-specific identifier (password hash, public ATProto DID, google email, etc.)
```
arushibandi: can you double check if we should explicitly make LoginID and Role `gorm:non null`

internal/org/models.go
```diff
@@ -6,17 +6,18 @@ import "time"
...
 
 // Keep track of members in the org.
-// Each member belongs to exactly one org.
-type member struct {
-	OrgID        string `gorm:"primaryKey"`
-	Member       string `gorm:"primaryKey"`
-	Role         string
-	PasswordHash string `gorm:"not null"`
+// Each Member belongs to exactly one org.
+type Member struct {
```
arushibandi: why did this become exported? for the most part the gorm model should remain unexported to avoid leaking impl details

internal/oauthserver/oauth_server.go
```diff
@@ -42,10 +42,10 @@ const (
 // This data is temporarily stored during the OAuth authorization flow to preserve
 // request context across redirects.
 type authRequestFlash struct {
-	Form          url.Values         // Original authorization request form data
-	ProviderType  login.ProviderType // Which login provider initiated this flow
-	ProviderState []byte             // Opaque provider-specific state
-	Did           syntax.DID         // DID of the user
+	Form          url.Values // Original authorization request form data
+	LoginMethod   string     // Which login method initiated this flow
```
arushibandi: yeah use the `LoginMethod` type here.

internal/login/provider.go
```diff
@@ -8,58 +8,43 @@ import (
...
 type Provider interface {
-	// Type returns a stable identifier stored in the session flash so
-	// HandleCallback can route to the correct provider without re-resolving the DID.
-	Type() ProviderType
-
-	// CanHandle returns true if this provider handles the given identity.
-	CanHandle(id *identity.Identity) bool
+	// LoginMethod returns a stable identifier used to route authorization
+	// requests to the correct provider based on the org's configured method.
+	LoginMethod() string
```
arushibandi: use it here too

internal/login/provider.go
```diff
@@ -8,58 +8,43 @@ import (
...
-// For returns the first provider whose CanHandle returns true.
-// It is assumed that only one provider matches the given identity (for now)
-func (r *Router) For(id *identity.Identity) (Provider, error) {
-	for _, p := range r.providers {
-		if p.CanHandle(id) {
-			return p, nil
-		}
+	r := &Router{providers: make(map[string]Provider, len(providers))}
+	for _, p := range providers {
+		r.providers[p.LoginMethod()] = p
```
arushibandi, line 43: should probs do a bit of checking here to make sure we fail if there is duplicates

internal/login/pds.go
```diff
@@ -26,19 +27,26 @@ type pdsProviderState struct {
...
-	_, hasHabitat := id.Services["habitat"]
-	return hasPDS && !hasHabitat
-}
+func (p *pdsProvider) Authorize(ctx context.Context, id *identity.Identity, loginID string) (string, []byte, error) {
+	// If the member has a public ATProto DID as their loginID, resolve it and use
+	// that identity's PDS for the OAuth flow. If no loginID (e.g. everyone org),
+	// use the identity as-is.
+	if loginID != "" && p.dir != nil {
+		publicDID, err := syntax.ParseDID(loginID)
+		if err == nil {
```
arushibandi: should we do `if err != nil { return err }` here and below ?

internal/oauthserver/oauth_server.go
```diff
@@ -284,13 +284,29 @@ func (o *OAuthServer) HandleAuthorize(
...
 		return
 	}
-	redirect, providerState, err := provider.Authorize(ctx, id)
+
+	// Look up the member's login ID (provider-specific identifier) from the store.
+	// If the DID isn't a member (e.g. everyone org), loginID stays empty.
+	loginID := ""
+	if member, err := o.orgStore.GetMember(ctx, id.DID); err == nil {
+		loginID = member.LoginID
+	}
```
arushibandi, line 329: should return if `err != nil` ?

internal/org/org.go
```diff
@@ -119,10 +106,18 @@ func NewOrg(
 	}, nil
 }
 
+func (s *orgImpl) LoginMethod() string {
+	var org organization
+	if err := s.db.First(&org, "id = ?", s.orgID).Error; err != nil {
+		return "password" // safe default
```
arushibandi, line 112: shouldn't this be an error ?

arushibandi: have `LoginMethodUnknown` and then have the caller return an error if you want to keep the types the same here.

internal/org/store.go
```diff
@@ -0,0 +1,176 @@
...
+var _ Store = &storeImpl{}
+
+// NewStore creates a Store that manages multiple orgs on a pear instance.
+func NewStore(
+	db *gorm.DB,
+	hve hive.Hive,
+	dir identity.Directory,
+	pearDomain string,
+) (Store, error) {
+	if err := db.AutoMigrate(&organization{}, &Member{}, &spentToken{}); err != nil {
```
arushibandi: i think we migrated &Member{} and &spentToken{} somewhere else too. so this should just do organization{}

internal/org/store.go
```diff
@@ -0,0 +1,176 @@
...
+		everyone:   NewEveryoneOrg(),
+	}, nil
+}
+
+func (s *storeImpl) orgFromModel(org *organization) (*orgImpl, error) {
+	signingSecret, err := base64.StdEncoding.DecodeString(org.SigningSecret)
+	if err != nil {
+		return nil, err
+	}
+	return &orgImpl{
```
arushibandi, line 74: cache this for perf? seems unnecessary to create a new &orgImpl each time. could just keep a `sync.Map`

internal/org/public.go
```diff
@@ -17,6 +17,10 @@ var (
 	ErrNotSupportedPublic = errors.New("method not supported on public org")
 )
 
+func (e *everyoneOrg) LoginMethod() string {
+	return "atproto"
```
arushibandi: nice

arushibandi (approved):

internal/org/models.go
sashankg: we need to be able to get members from other packages, right?

internal/org/store.go
sashankg: no i just moved the store into its own file

internal/org/store.go
sashankg: we're just allocating memory, right? i don't think we need to worry about the perf of that

PR #322, opened 2026-05-15
[Caddy] Support multilevel

sashankg:

### DIFF
--- a/Caddyfile
+++ b/Caddyfile
@@ -3,7 +3,7 @@ frontend.local.habitat.network {
 	tls internal
 }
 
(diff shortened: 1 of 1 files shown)


arushibandi (approved):

arushibandi:
why do we need this tho

sashankg:
i hooked up `*.id.<DOMAIN>` to be the default hiveDomain. i can override it locally to be `*.<DOMAIN>` but thought it would be good to keep separate since it matches prod and also we'll be able to catch issues where we might be using the wrong domain

PR #325, opened 2026-05-16
[Org] Add auth method to setup

sashankg:
Screenshot 2026-05-15 at 6 21 10 PM
Screenshot 2026-05-15 at 6 21 18 PM

### DIFF
--- a/cmd/pear/main.go
+++ b/cmd/pear/main.go
@@ -128,9 +128,9 @@ func run(_ context.Context, cmd *cli.Command) error {
 
 	domain := cmd.String(fDomain)
 	oauthClient, err := pdsclient.NewPdsOAuthClient(
-		"https://sashankg.github.io/client-metadata.json", /*clientId*/
-		"https://"+domain,                   /*clientUri*/
-		"https://"+domain+"/oauth-callback", /*redirectUri*/
+		"https://"+domain+"/client-metadata.json", /*clientId*/
+		"https://"+domain,                         /*clientUri*/
+		"https://"+domain+"/oauth-callback",       /*redirectUri*/
 		cmd.String(fOauthClientSecret),
 		meter,
 	)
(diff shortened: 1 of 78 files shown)


cmd/pear/main.go
```diff
@@ -128,9 +128,9 @@ func run(_ context.Context, cmd *cli.Command) error {
 
 	domain := cmd.String(fDomain)
 	oauthClient, err := pdsclient.NewPdsOAuthClient(
-		"https://"+domain+"/client-metadata.json", /*clientId*/
-		"https://"+domain,                         /*clientUri*/
-		"https://"+domain+"/oauth-callback",       /*redirectUri*/
+		"https://sashankg.github.io/client-metadata.json", /*clientId*/
```
arushibandi: remove

internal/login/pds.go
```diff
@@ -27,24 +27,34 @@ type pdsProviderState struct {
...
+	dir identity.Directory,
+) Provider {
 	return &pdsProvider{oauthClient: oauthClient, credStore: credStore, dir: dir}
 }
 
 func (p *pdsProvider) LoginMethod() string { return "atproto" }
 
-func (p *pdsProvider) Authorize(ctx context.Context, id *identity.Identity, loginID string) (string, []byte, error) {
+func (p *pdsProvider) Authorize(
+	ctx context.Context,
```
arushibandi, line 42: can we add a go formatter so these changes don't happen randomly? i don't have the same formatter  / settings installed as you

internal/org/server_test.go
```diff
@@ -29,7 +29,7 @@ func newTestServer(t *testing.T, adminDID syntax.DID) (*Server, string) {
 	storeImpl, err := NewStore(db, h, identity.DefaultDirectory(), "pear.example.com")
 	require.NoError(t, err)
 
-	orgId, _, err := storeImpl.CreateOrg(t.Context(), "test-org", "admin", "password")
+	orgId, _, err := storeImpl.CreateOrg(t.Context(), "test-org", "admin", "password", "", "")
```
arushibandi: should probably be adding new tests for the new login methods right?

arushibandi (approved):
LGTM can you also make sure the form takes in handle rather than just DIDs for the atproto login type ?

PR #326, opened 2026-05-16
[Auth] Sign in with google

sashankg:


internal/login/google.go
```diff
@@ -1,27 +1,156 @@
...
+
+	authURL := p.oauthCfg.AuthCodeURL(
+		stateStr,
+		oauth2.AccessTypeOffline,
+		oauth2.S256ChallengeOption(verifier),
+		oauth2.SetAuthURLParam("login_hint", loginID),
+		oauth2.SetAuthURLParam("prompt", "select_account"),
+	)
+
+	return authURL, stateBytes, nil
```
arushibandi, line 98: just thinking ahead ... at some point we'll want the org to verify certain domains / emails as safe to join the org.

arushibandi: maybe that's the same as the admin adding them by email... in which case i apologize

arushibandi (approved):

PR #330, opened 2026-05-19
[Hive] Support Habitat-Host header

sashankg:
GCP doesn't support mismatched hostnames. this also seems like a useful option in the future for others who want to host hive similarly

### DIFF
--- a/internal/hive/server.go
+++ b/internal/hive/server.go
@@ -8,6 +8,17 @@ import (
 	"github.com/bluesky-social/indigo/atproto/syntax"
 )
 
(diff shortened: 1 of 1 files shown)


arushibandi (approved):

arushibandi:
wdym mismatched host names

sashankg:
We put a reverse proxy in front of gcp (on railway) to catch all subdomains. If we try to keep the hostname the same as the original request gcp complains. We can get the reverse proxy to map the original host to this new header as a simple fix.

PR #331, opened 2026-05-19
[Org] Support subdomains

sashankg:

### DIFF
--- a/internal/hive/hive.go
+++ b/internal/hive/hive.go
@@ -113,20 +113,17 @@ func (h *hive) LookupDID(ctx context.Context, did syntax.DID) (*identity.Identit
 	if !ok {
 		return nil, identity.ErrDIDNotFound
 	}
-
 	return h.store.getIdentityByID(ctx, opaqueID)
 }
 
 // LookupHandle implements identity.Directory
-// It strips the member domain suffix from the given handle. Handle format is
-// <internal-handle>.<memberDomain> where <internal-handle> may contain dots
+// It strips the member domain suffix from the given handle. Handle format is <internal-handle>.<memberDomain>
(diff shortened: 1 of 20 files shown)


internal/hive/hive.go
```diff
@@ -118,18 +118,14 @@ func (h *hive) LookupDID(ctx context.Context, did syntax.DID) (*identity.Identit
 }
 
 // LookupHandle implements identity.Directory
-// It strips the internal handle prefix from the given handle which has format
-// <internal-handle>.membersNamespace.domain or <internal-handle>.domain if membersNamespace == ""
-// and looks up the handle against the store.
+// It strips the member domain suffix from the given handle. Handle format is
+// <internal-handle>.<memberDomain> where <internal-handle> may contain dots
```
arushibandi: why does it contain dots?

internal/org/models.go
```diff
@@ -16,11 +16,12 @@ const (
...
-	Name          string      // optional display name
-	LoginMethod   LoginMethod // "atproto", "google", "password"
-	SigningSecret string      // base64-encoded HMAC-SHA256 key for invite tokens
-	CreatedAt     time.Time
+	ID              string      `gorm:"primaryKey"`
+	Name            string      // optional display name
+	LoginMethod     LoginMethod // "atproto", "google", "password"
+	SigningSecret   string      // base64-encoded HMAC-SHA256 key for invite tokens
+	CreatedAt       time.Time
+	HandleSubdomain string `gorm:"unique"`
```
arushibandi, line 24: eventually this should become mutable ... like you can start with my-org.habitat.network but then move to my-org.com if you end up self-hosting ... but fine for now i think

arushibandi (approved):

PR #333, opened 2026-05-19
[Migrations] Setup goose

sashankg:


cmd/pear/migrations/20260519200821_init.go
```diff
@@ -0,0 +1,21 @@
...
+
+func upInit(ctx context.Context, tx *sql.Tx) error {
+	// This code is executed when the migration is applied.
+	return nil
+}
+
+func downInit(ctx context.Context, tx *sql.Tx) error {
+	// This code is executed when the migration is rolled back.
+	return nil
+}
```
arushibandi, line 21: why not just call it `up` and `down`

arushibandi (approved):
nice

cmd/pear/migrations/20260519200821_init.go
sashankg: there will be multiple up and downs in this package for each migration. the method names are also generated by `moonx pear:migration-create -- init go`. since the name of the migration is init, that's what it calls the methods

PR #338, opened 2026-05-20
[Spaces] Basic implementation

sashankg:


internal/spaces/spaces.go
```diff
@@ -0,0 +1,434 @@
...
+	"time"
+
+	"github.com/bluesky-social/indigo/atproto/syntax"
+	"github.com/google/uuid"
+	"gorm.io/gorm"
+	"gorm.io/gorm/clause"
+)
+
+// SpaceURI identifies a space.
+// Format: "habitat://spaceDID/spaceType/skey"
```
arushibandi: do we need the `habitat://` here? i think not any more (we were doing that for private records to not conflate with existing at records but seems like we are compatible with space URIs by default if all spaces are owned by the org and the org has a DID ?)

internal/spaces/spaces.go
```diff
@@ -0,0 +1,434 @@
...
+}
+
+// MemberInfo holds a member's DID and when they were added
+type MemberInfo struct {
+	Did     syntax.DID
+	AddedAt time.Time
+}
+
+// RecordInSpace is a single record within a space
+type RecordInSpace struct {
```
arushibandi: just call this `Record`
it becomes `spaces.Record` which is more readable than `spaces.RecordInSpace`

internal/spaces/spaces.go
```diff
@@ -0,0 +1,434 @@
...
+	Skey      string `gorm:"primaryKey"`
+	Type      string `gorm:"index"`
+	CreatedAt time.Time
+}
+
+// TODO: members table will be added when the permission store is built.
+// For now, the owner is always the sole member of a space.
+
+type spaceRecord struct {
+	Owner      string `gorm:"primaryKey"`
```
arushibandi: does this need to be multi-tenant w.r.t orgs?

internal/spaces/spaces.go
```diff
@@ -0,0 +1,434 @@
...
+type SpaceView struct {
+	URI         SpaceURI
+	Type        syntax.NSID
+	Skey        string
+	MemberCount int
+}
+
+// MemberInfo holds a member's DID and when they were added
+type MemberInfo struct {
+	Did     syntax.DID
```
arushibandi, line 123: doesn't this need to be keyed by space? i don't see it used in the other models also

internal/spaces/spaces.go
```diff
@@ -0,0 +1,434 @@
...
+		return ""
+	}
+	nsid, err := syntax.ParseNSID(parts[2])
+	if err != nil {
+		return ""
+	}
+	return nsid
+}
+
+func (s SpaceURI) Skey() string {
```
arushibandi: can we formally type this too? (even if its just a string). eventually it will probs become `syntax.SpaceKey`

internal/spaces/spaces.go
```diff
@@ -0,0 +1,434 @@
...
+	err := s.db.WithContext(ctx).
+		Where("owner = ? AND skey = ?", uri.SpaceDID().String(), uri.Skey()).
+		First(&sp).Error
+	if errors.Is(err, gorm.ErrRecordNotFound) {
+		return false, nil
+	} else if err != nil {
+		return false, err
+	}
+
+	return did == uri.SpaceDID(), nil
```
arushibandi, line 308: i don't get this
it's not using the returned space

internal/spaces/spaces.go
```diff
@@ -0,0 +1,434 @@
...
+	if collection != nil {
+		query = query.Where("collection = ?", collection.String())
+	}
+
+	var rows []spaceRecord
+	if err := query.Order("rkey ASC").Find(&rows).Error; err != nil {
+		return nil, err
+	}
+
+	records := make([]RecordInSpace, len(rows))
```
arushibandi: add a todo around cursors here (i thin k the lexicon included that)

internal/spaces/spaces.go
sashankg: yeah this should be `ats://`

internal/spaces/spaces.go
sashankg: owner is a did which is unique across orgs.

internal/spaces/spaces.go
sashankg: yeah this will fetched from the permission store. rn we just return the owner as the only member for every space

internal/spaces/spaces.go
sashankg: we're just returning the space owner as the only member for now. we need the permission store to return the actual members

arushibandi (approved):

PR #339, opened 2026-05-21
[Spaces] Connect to fgastore

sashankg:
-------------------------
- master
  - **[Spaces] Connect to fgastore** :point_left:
    - #342
      - #346
      - #345


internal/fgastore/encoding.go
```diff
@@ -0,0 +1,52 @@
...
+
+	"github.com/bluesky-social/indigo/atproto/syntax"
+
+	habitat_syntax "github.com/habitat-network/habitat/internal/syntax"
+)
+
+// OpenFGA uses ":" as a delimiter in user and object references,
+// but DIDs contain ":" (e.g. "did:plc:alice").  We use "_" as a
+// stand-in for ":" when encoding DIDs into FGA tuple strings, and
+// reverse the substitution when parsing them back.
```
arushibandi: ☹️

internal/fgastore/encoding.go
```diff
@@ -0,0 +1,52 @@
...
+}
+
+func fgaDecodeDID(s string) (syntax.DID, error) {
+	return syntax.ParseDID(strings.ReplaceAll(s, "_", ":"))
+}
+
+// SpaceObjectKey returns the FGA object key for a space.
+// The key is the full SpaceURI with colons replaced by underscores
+// so OpenFGA can parse it as a typed object.
+func SpaceObjectKey(uri habitat_syntax.SpaceURI) string {
```
arushibandi, line 19: can you add some documentation about what the FGA object key looks like and maybe make that a named type as well?

internal/fgastore/fgastore.go
```diff
@@ -16,12 +16,31 @@ import (
 
 var _ Store = (*FGA)(nil)
 
+// Tuple represents a relationship tuple (used as contextual tuples in requests).
+type Tuple struct {
```
arushibandi, line 20: why do we have to define this type ourselves? i would've thought the fga library wouldve given it

internal/spaces/spaces.go
```diff
@@ -5,116 +5,40 @@ import (
...
-
 type spaceRecord struct {
-	Space      SpaceURI    `gorm:"primaryKey"`
-	Owner      syntax.DID  `gorm:"primaryKey"`
-	Collection syntax.NSID `gorm:"primaryKey"`
-	Rkey       string      `gorm:"primaryKey"`
+	Space      habitat_syntax.SpaceURI `gorm:"primaryKey"`
+	Owner      syntax.DID              `gorm:"primaryKey"`
+	Collection syntax.NSID             `gorm:"primaryKey"`
+	Rkey       string                  `gorm:"primaryKey"`
```
arushibandi: `syntax.Rkey?`

internal/spaces/spaces.go
```diff
@@ -239,32 +184,76 @@ func (s *store) ListSpaces(
...
+		if err := s.db.WithContext(ctx).Model(&space{}).
+			Where("owner = ?", actor).
+			Order("created_at DESC").
+			Find(&rows).Error; err != nil {
+			return nil, err
+		}
 
-	if filterType != nil {
-		query = query.Where("type = ?", filterType.String())
+		fgaObjects, err := s.fga.ListObjects(
```
arushibandi: can you add some comments in this function

arushibandi: about what this is listing -- seems to be all of the actor's spaces they are in, but the code is just confusing to read a bit

arushibandi: talked irl, fix the type so it only takes in one did

arushibandi (approved):

internal/fgastore/encoding.go
sashankg: using url encoding instead

internal/fgastore/encoding.go
sashankg: yeah will be doing in a follow up

internal/fgastore/fgastore.go
sashankg: once we define our own types for what the possible values are this will be more useful

internal/spaces/spaces.go
sashankg: oh i didn't know that was a thing

PR #346, opened 2026-05-21
[Spaces] FGAStore migrations

sashankg:


internal/fgastore/fgastore.go
```diff
@@ -51,15 +55,68 @@ type FGA struct {
...
+	)
+	if err != nil {
+		return nil, fmt.Errorf("fgastore migration provider: %w", err)
+	}
+	if _, err := provider.Up(ctx); err != nil {
+		return nil, fmt.Errorf("fgastore sqlite migrations: %w", err)
+	}
+	ds, err := sqlite.New(uri, &sqlcommon.Config{
+		MaxTuplesPerWriteField: 100,
+		MaxTypesPerModelField:  100,
```
arushibandi: what's this

arushibandi (approved):

internal/fgastore/fgastore.go
sashankg: the defaults. switched to using the constants from openfga

PR #352, opened 2026-05-22
[Cliques] Migration script

sashankg:


cmd/migrate-cliques/main.go
```diff
@@ -0,0 +1,87 @@
...
+		Flags: []cli.Flag{
+			&cli.StringFlag{
+				Name:  "pg-url",
+				Usage: "Postgres connection URL",
+			},
+			&cli.StringFlag{
+				Name:  "db",
+				Usage: "SQLite database path",
+			},
+		},
```
arushibandi, line 31: are these supposed to be mutually exclusive

cmd/migrate-cliques/migrate.go
```diff
@@ -0,0 +1,177 @@
...
+	ctx context.Context,
+	db *gorm.DB,
+	cliques CliqueMembersReader,
+	spaces SpaceWriter,
+) (int, error) {
+	rows, err := db.Raw(`
+		SELECT DISTINCT grantee FROM permissions
+		WHERE collection IN (?, ?)
+		AND grantee LIKE ?
+	`, "network.habitat.docs", "network.habitat.docs.edit", "clique:%").Rows()
```
arushibandi, line 53: check if there are any docs that have both direct DID grantee + clique (or just direct grants) since these will break now

cmd/migrate-cliques/migrate.go
```diff
@@ -0,0 +1,177 @@
...
+	}
+	suffix := ""
+	if db.Dialector.Name() == "postgres" {
+		suffix = " ON CONFLICT (space, owner, collection, rkey) DO NOTHING"
+	}
+	stmt := fmt.Sprintf(`
+		%s space_records (space, owner, collection, rkey, value, created_at, updated_at)
+		SELECT ? AS space, p.owner, p.collection, p.rkey, r.value,
+		       COALESCE(r.created_at, CURRENT_TIMESTAMP),
+		       COALESCE(r.updated_at, CURRENT_TIMESTAMP)
```
arushibandi, line 162: can we preserve these ?

cmd/migrate-cliques/migrate.go
arushibandi: also check if we should not filter by collection and migrate everything.

arushibandi (approved):

PR #353, opened 2026-05-22
[Spaces] Fix ListSpaces impl

sashankg:
I was right the first time. ListSpaces lists all the spaces the caller is a member of. you can optionally filter by space owner

### DIFF
--- a/internal/spaces/spaces.go
+++ b/internal/spaces/spaces.go
@@ -187,6 +187,7 @@ func (s *store) ListSpaces(
 	filterType *syntax.NSID,
 ) ([]SpaceView, error) {
 	var allRows []space
+	// start with owned spaces
 	ownedRowsQuery := s.db.WithContext(ctx).Model(&space{}).Where("owner = ?", member)
 	if filterOwner != nil {
 		ownedRowsQuery = ownedRowsQuery.Where("owner = ?", *filterOwner)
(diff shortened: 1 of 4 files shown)


internal/spaces/spaces.go
```diff
@@ -179,27 +181,100 @@ func (s *store) CreateSpace(
 
 func (s *store) ListSpaces(
 	ctx context.Context,
-	owner syntax.DID,
+	member syntax.DID,
+	filterOwner *syntax.DID,
```
arushibandi, line 186: comment on this function because it's confusing

internal/spaces/spaces.go
```diff
@@ -179,27 +181,100 @@ func (s *store) CreateSpace(
...
-	var rows []space
+	type spaceKey struct {
+		owner string
+		skey  string
+	}
+	seen := make(map[spaceKey]struct{})
+	var allKeys []spaceKey
+
+	var ownedRows []space
+	if err := s.db.WithContext(ctx).Model(&space{}).
```
arushibandi: comments throughout this function would be appreciated too, just bc it's a bit confusing

internal/spaces/spaces.go
```diff
@@ -179,27 +181,100 @@ func (s *store) CreateSpace(
...
+	)
+	if err != nil {
+		return nil, fmt.Errorf("list fga member spaces: %w", err)
+	}
+	for _, obj := range fgaObjects {
+		uri, err := fgastore.ParseSpaceObjectKey(obj)
+		if err != nil {
+			continue
+		}
+		key := spaceKey{owner: uri.SpaceDID().String(), skey: string(uri.Skey())}
```
arushibandi: maybe have a little constructor for spacekey?

arushibandi (approved):

internal/spaces/spaces.go
sashankg: simplified the function so it's easier to read now and removed `spaceKey`

PR #356, opened 2026-05-23
[DB] Support cross store transactions

sashankg:


internal/org/org.go
```diff
@@ -331,22 +321,19 @@ func (s *orgImpl) CreateNewMemberIdentity(
...
-	if err != nil {
-		return nil, err
-	}
-
+	var id *identity.Identity
 	err = s.db.WithContext(ctx).Transaction(func(tx *gorm.DB) error {
-		if err := persistIdent(tx); err != nil {
-			return err
+		// If token is valid, call into hive to mint the new identity and serve it
+		newId, err := db.WithTransaction(tx, s.hive).MintIdentity(internalHandle, s.handleSubdomain)
```
arushibandi: instead of this can you change it so the syntax is `s.hive.WithTransaction(tx).MintIdentity(...)`

I think you can just call the `WithTx` function you wrote? and delete `db.WithTransaction`?

sashankg: yeah i'm on the fence. that would leak the db implementation detail of hive to the interface. but maybe that's too pedantic and implementations that don't support transactions can just be noops. the current approach is definitely not very explicit

arushibandi: ya if they don't support transactions it can be a noop

arushibandi (approved):

PR #363, opened 2026-05-26
[Spaces] Add access to API

sashankg:

### DIFF
--- a/typescript/api/lexicons.ts
+++ b/typescript/api/lexicons.ts
@@ -2471,6 +2471,13 @@ export const schemaDict = {
                 format: 'did',
                 description: 'The DID of the user to add.',
               },
+              access: {
+                type: 'string',
+                enum: ['read', 'write'],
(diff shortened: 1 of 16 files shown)


typescript/api/lexicons.ts
```diff
@@ -2655,6 +2662,10 @@ export const schemaDict = {
             type: 'string',
             format: 'did',
           },
+          access: {
+            type: 'string',
+            enum: ['read', 'write'],
+          },
```
arushibandi, line 2668: is this supported natively in spaces?

arushibandi (approved):

typescript/api/lexicons.ts
sashankg: he mentions it in https://dholms.leaflet.pub/3mhj6bcqats2o#access-control but it's missing in the draft implementation. ultimately there does need to be something like this so adding it for now

PR #364, opened 2026-05-27
[Auth] Org scope

sashankg:
- **feat(oauth): add permission scope types, parsing, and matching**
- **fixup: remove repo from validResources, use Permission structs directly in tests**
- **feat(oauth): enforce admin-only org scope grant and custom scope validation**
- **fixup: address code review - use strings.HasPrefix, add admin gate test, fix duplicate require**
- **fixup: pass nil auth to NewServer in jetstream tests**
- **fixup: add strings.HasPrefix guard in HandleCallback, fix TokenValidator return type to syntax.DID**
- **fixes**


internal/jetstream/server.go
```diff
@@ -8,41 +8,42 @@ import (
...
-// This will be added as a follow up.
-
-// The Server handles listening for updates from an ingestion stream and fanning them out to many upstream
-// jetstream subscribers over HTTP SSE. We don't use Websocket because it's more expensive and complex operationally,
-// and this usecase doesn't support the client sending messages.
-//
-// The official jetstream API uses bi-directionality so that the client can update
-// what collections / repos / etc. it is subscribed to. This is a simple enough usecase
-// that the client can close the connection and re-open it with the new parameters and the
-// timestamp cursor from where it left off.
```
arushibandi, line 28: nooo don't delete my comments!!

internal/oauthserver/permission.go
```diff
@@ -0,0 +1,133 @@
+package oauthserver
+
+import (
+	"fmt"
+	"net/url"
+	"strings"
+)
+
+var validResources = map[string]bool{
+	"org": true,
```
arushibandi: nit: typed constants always

internal/oauthserver/permission.go
```diff
@@ -0,0 +1,133 @@
...
+}
+
+type permission struct {
+	Resource   string
+	Collection string
```
arushibandi: can we make this `syntax.NSID`?

internal/oauthserver/permission.go
```diff
@@ -0,0 +1,133 @@
...
+	Actions    []string
```
arushibandi: same thing typed constants here? `[]permissionAction`

internal/oauthserver/permission.go
```diff
@@ -0,0 +1,133 @@
...
+	"net/url"
+	"strings"
+)
+
+var validResources = map[string]bool{
+	"org": true,
+}
+
+type permission struct {
+	Resource   string
```
arushibandi: same thing

internal/oauthserver/permission.go
```diff
@@ -0,0 +1,133 @@
...
+	if granted.Collection != "*" && granted.Collection != required.Collection {
+		return false
+	}
+	if len(required.Actions) == 0 {
+		return true
+	}
+	if len(granted.Actions) == 0 {
+		return true
+	}
+	requiredSet := make(map[string]bool, len(required.Actions))
```
arushibandi: you can do `xmaps.Set` / make it `map[string]struct{}` for a set (so there's no true | false | existence triple)

internal/oauthserver/permission.go
```diff
@@ -0,0 +1,133 @@
...
+	for _, a := range required.Actions {
+		requiredSet[a] = true
+	}
+	for _, a := range granted.Actions {
+		delete(requiredSet, a)
```
arushibandi: there might be some kind of `maps.` helper to use here even

internal/oauthserver/permission.go
```diff
@@ -0,0 +1,133 @@
...
+	}
+	return len(requiredSet) == 0
+}
+
+func scopesSatisfy(grantedScopes, requiredScopes []string) bool {
```
arushibandi: i think i'm blind but where does this get used?

arushibandi (approved):

PR #365, opened 2026-05-27
[debug traces]: log in prod if the tracer is working

arushibandi:


cmd/pear/main.go
```diff
@@ -168,6 +168,15 @@ func run(_ context.Context, cmd *cli.Command) error {
 	// Order of middlewares = order of "Use" called
 	// https://pkg.go.dev/go.opentelemetry.io/contrib/instrumentation/github.com/gorilla/mux/otelmux
 	mux.Use(otelmux.Middleware("habitat-server"))
+	mux.Use(func(next http.Handler) http.Handler {
+		return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+			_, span := otel.Tracer("diagnostic").Start(r.Context(), "diagnostic-span")
+			log.Info().Bool("is_recording", span.IsRecording()).Msg("diagnostic span check")
+			span.End()
```
sashankg: do you not want the span to end after handling the request?

sashankg (approved):

cmd/pear/main.go
arushibandi: this was just creating a mini span at the beginning but i removed it to try a different thing

PR #366, opened 2026-05-28
[Spaces] Delete space

sashankg:


internal/spaces/server.go
```diff
@@ -594,3 +594,47 @@ func (s *Server) DeleteRecord(w http.ResponseWriter, r *http.Request) {
...
+		return
+	}
+
+	spaceURI, err := habitat_syntax.ParseSpaceURI(input.Space)
+	if err != nil {
+		utils.LogAndHTTPError(w, err, "parse space uri", http.StatusBadRequest)
+		return
+	}
+
+	authorized, err := s.authorize(r.Context(), callerDID, spaceURI, "owner")
```
arushibandi: again would be nice to not just have strings and have `spaces.Owner` or something here

internal/spaces/spaces.go
```diff
@@ -567,6 +568,58 @@ func (s *store) ListRecords(
...
+	// verify space exists before starting the transaction
+	var sp space
+	err := s.db.WithContext(ctx).
+		Where("owner = ? AND skey = ?", uri.SpaceDID(), uri.Skey()).
+		First(&sp).Error
+	if errors.Is(err, gorm.ErrRecordNotFound) {
+		return ErrSpaceNotFound
+	} else if err != nil {
+		return err
+	}
```
arushibandi: i think we can do one less query by removing this and below just doing `First().Delete()` and checking ErrRecordNotFOund there?

arushibandi (approved):

PR #370, opened 2026-05-28
[Spaces] Oplog

sashankg:


internal/spaces/spaces_test.go
```diff
@@ -463,3 +463,65 @@ func TestDeleteSpace_NonExistent(t *testing.T) {
 	err := s.DeleteSpace(t.Context(), uri)
 	require.ErrorIs(t, err, ErrSpaceNotFound)
 }
+
+func TestGetRepoOplog(t *testing.T) {
```
arushibandi, line 467: can you add a test around sorting and the use of TIDs?

internal/spaces/spaces.go
```diff
@@ -22,9 +22,10 @@ import (
 
 type space struct {
 	Owner     syntax.DID              `gorm:"primaryKey"`
+	Type      syntax.NSID             `gorm:"primaryKey"`
 	Skey      habitat_syntax.SpaceKey `gorm:"primaryKey"`
-	Type      syntax.NSID
 	CreatedAt time.Time
+	Rev       syntax.TID `gorm:"uniqueIndex"`
```
arushibandi, line 28: do we need to put another index on this to be able to sort?

arushibandi (approved):

internal/spaces/spaces.go
sashankg: it'll make the sorting faster, right?

PR #371, opened 2026-05-30
[Spaces] Add CIDs

sashankg:

### DIFF
--- a/internal/syntax/space.go
+++ b/internal/syntax/space.go
@@ -93,3 +93,18 @@ func (s SpaceURI) Skey() SpaceKey {
 func (s SpaceURI) String() string {
 	return string(s)
 }
+
+type SpaceRecordURI string
+
+func ConstructSpaceRecordURI(
+	spaceUri SpaceURI,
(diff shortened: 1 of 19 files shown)


internal/syntax/space.go
```diff
@@ -93,3 +93,18 @@ func (s SpaceURI) Skey() SpaceKey {
...
+
+type SpaceRecordURI string
+
+func ConstructSpaceRecordURI(
+	spaceUri SpaceURI,
+	repo syntax.DID,
+	collection syntax.NSID,
+	rkey syntax.RecordKey,
+) SpaceRecordURI {
+	return SpaceRecordURI(fmt.Sprintf("%s/%s/%s/%s", spaceUri, repo, collection, rkey))
```
arushibandi, line 105: doesn't this need `ats://`

arushibandi (approved):

internal/syntax/space.go
sashankg: that's already in the spaceUri

PR #376, opened 2026-06-02
small ui fix for login header

arushibandi:


sashankg:
this shouldn't be necessary. if the authManager fails because of `UnauthenticatedError` it will trigger the logout which invalidates the router state. the loader will run again and have no authInfo.

PR #382, opened 2026-06-06
[Auth] Refactor providers to support org auth

sashankg:
Move the routing logic into org package and make `loginMethod` internal detail. Will be extending the LoginRouter to handle authenticating on behalf of the entire org.


cmd/pear/main.go
```diff
@@ -207,17 +200,33 @@ func run(_ context.Context, cmd *cli.Command) error {
...
+	passwordProvider, err := login.NewPasswordProvider(
+		db,
 		cmd.String(fDomain),
 		cmd.String(fFrontendDomain),
 		oauthSecret,
 		dir,
 	)
+	if err != nil {
+		slog.Error("unable to setup password login provider", "err", err)
+		os.Exit(1)
```
arushibandi, line 212: does slog not have a `Fatal` log level?

internal/org/login_router.go
```diff
@@ -0,0 +1,102 @@
...
+	// member login
+	if err == nil {
+		provider := r.getProvider(member.Org)
+		if provider == nil {
+			return "", nil, fmt.Errorf("unsupported login provider for %s", did)
+		}
+		return provider.Authorize(ctx, member.LoginID)
+	} else if !errors.Is(err, ErrMemberNotFound) {
+		return "", nil, fmt.Errorf("failed to get member: %w", err)
+	}
```
arushibandi, line 45: I think this is wrong
```
if errors.Is(err, ErrMemberNotFound) {
    return "", nil, fmt.Errorf("failed to get member: %w", err)
} else if err != nil {
    return "", nil. err
}

provider := r.getProvider(member.Org)
if provider == nil {
	return "", nil, fmt.Errorf("unsupported login provider for %s", did)
}
return provider.Authorize(ctx, member.LoginID)
```

internal/org/org.go
```diff
@@ -60,8 +60,7 @@ type Org interface {
 	IsAdmin(ctx context.Context, did syntax.DID) (bool, error)
 	IsMember(ctx context.Context, did syntax.DID) (bool, error)
 
-	// LoginMethod returns how users authenticate: "atproto", "google", or "password".
-	LoginMethod(ctx context.Context) LoginMethod
+	loginMethod() loginMethod
```
arushibandi, line 63: again, don't remove comments

arushibandi (approved):

cmd/pear/main.go
sashankg: unfortunately no. will revisit to make a helper function or something

internal/org/login_router.go
sashankg: no we're intentionally falling through if there is no member with that did and checking if the did is the org id

internal/org/org.go
sashankg: this is internal now hence removed the comment. still have a comment on the model itself

PR #383, opened 2026-06-07
[Sync] Event store

sashankg:


internal/sync/server.go
```diff
@@ -0,0 +1,89 @@
...
+	cursor := r.URL.Query().Get("cursor")
+	var lastSeq uint64
+	if cursor != "" {
+		parsed, err := strconv.ParseUint(cursor, 10, 64)
+		if err != nil {
+			utils.LogAndHTTPError(r.Context(), w, err, "parse cursor", http.StatusBadRequest)
+			return
+		}
+		lastSeq = parsed
+	} else if r.Header.Get("Last-Event-ID") != "" {
```
arushibandi, line 32: is this atproto syntax?

internal/sync/server.go
```diff
@@ -0,0 +1,89 @@
...
+type Server struct {
+	eventStore events.Store
+}
+
+func NewServer(eventStore events.Store) *Server {
+	return &Server{eventStore: eventStore}
+}
+
+func (s *Server) HandleSubscribeSpaces(w http.ResponseWriter, r *http.Request) {
+	cursor := r.URL.Query().Get("cursor")
```
arushibandi, line 23: auth ?

internal/events/store.go
```diff
@@ -0,0 +1,261 @@
...
+				}
+			}
+			s.subscribersMu.RUnlock()
+		}
+	}
+}
+
+// Subscribe implements [Store].
+func (s *storeImpl) Subscribe(ctx context.Context, since uint64) <-chan Event {
+	notificationCh := make(chan struct{}, 1)
```
arushibandi, line 184: should probably make a bigger buffer on this since it directly blocks the sequencer (what i was worried about)

internal/events/store.go
```diff
@@ -0,0 +1,261 @@
...
+				if err != nil {
+					slog.ErrorContext(ctx, "failed to get events", "err", err)
+					continue
+				}
+				for _, event := range events {
+					if event.Seq > lastSent {
+						select {
+						case <-ctx.Done():
+							return
+						case ch <- event:
```
arushibandi, line 215: eject subscribers if they can't keep up:
```
case <-ctx.Done():
return
case ch <- event:
lastSent = event.Seq
default:
close(ch)
```

internal/events/store.go
```diff
@@ -0,0 +1,261 @@
...
+				return
+			case <-notificationCh:
+				slog.DebugContext(ctx, "notifying subscriber")
+				events, err := s.GetEvents(ctx, lastSent)
+				if err != nil {
+					slog.ErrorContext(ctx, "failed to get events", "err", err)
+					continue
+				}
+				for _, event := range events {
+					if event.Seq > lastSent {
```
arushibandi, line 211: why do we need this? isn't it guaranteed?

internal/events/store.go
```diff
@@ -0,0 +1,261 @@
...
+				s.subscribersMu.Unlock()
+				return
+			case <-notificationCh:
+				slog.DebugContext(ctx, "notifying subscriber")
+				events, err := s.GetEvents(ctx, lastSent)
+				if err != nil {
+					slog.ErrorContext(ctx, "failed to get events", "err", err)
+					continue
+				}
+				for _, event := range events {
```
arushibandi, line 210: also should we be batching all of these together into one message?

internal/events/store.go
```diff
@@ -0,0 +1,261 @@
...
+	Type  string
+	Space habitat_syntax.SpaceURI
+	Time  time.Time
+	Repo  syntax.DID
+	Rev   syntax.TID `json:"rev,omitempty"`
+	Since syntax.TID `json:"since,omitempty"`
+	Ops   []EventOps
+}
+
+type EventOps struct {
```
arushibandi, line 38: shouldn't this be `EventOp`? a single operation?

internal/events/store.go
```diff
@@ -0,0 +1,261 @@
...
+	Action string
+
+	// RepoOp
+	Uri   habitat_syntax.SpaceRecordURI
+	Value map[string]any
+	Cid   string
+
+	// Membership
+	Did    syntax.DID `json:"did,omitempty"`
+	Access string
```
arushibandi, line 48: should be a named type?

internal/events/store.go
```diff
@@ -0,0 +1,261 @@
...
+type EventOps struct {
+	Action string
+
+	// RepoOp
+	Uri   habitat_syntax.SpaceRecordURI
+	Value map[string]any
+	Cid   string
+
+	// Membership
+	Did    syntax.DID `json:"did,omitempty"`
```
arushibandi, line 47: who is this? is the record writer not given by the space record URI?

internal/events/store.go
```diff
@@ -0,0 +1,261 @@
...
+// AppendSpaceEvent implements [Store].
+func (s *storeImpl) AppendSpaceEvent(
+	ctx context.Context,
+	space habitat_syntax.SpaceURI,
+	repo syntax.DID,
+	rev syntax.TID,
+	prev syntax.TID,
+	ops []EventOps,
+) error {
+	entry, err := json.Marshal(Event{
```
arushibandi, line 98: why json marshal rather than storing these directly? isn't that inefficient bc then the read path has to unmarshal?

internal/events/store_test.go
```diff
@@ -0,0 +1,80 @@
...
+				require.Greater(t, e.Rev, e.Since)
+				require.Greater(t, e.Seq, lastSeen)
+				lastSeen = e.Seq
+
+				if lastSeen >= totalEvents {
+					return
+				}
+			}
+		})
+	}
```
arushibandi, line 77: other tests:
- test that if 1 subscriber is blocked it doesn't block any others
- chaos test: have N subscribers, block some subset for random amount of time, when they unblock ensure that the updates they get are correct (this will make sure we don't have correctness regression or races as the code gets more complex)

internal/sync/server.go
sashankg: no it's SSE spec https://html.spec.whatwg.org/multipage/server-sent-events.html#the-last-event-id-header

internal/sync/server.go
sashankg: yeah will follow up after

internal/events/store.go
sashankg: nope this is just a notification channel. if it's already full, the sender just drops the notification because the receiver will query for all new messages after it receives the notifiactions

internal/events/store.go
sashankg: no i think it's fine to block the subscriber thread on the client. it's not blocking the sequencer nor other subscribers so it's fine. we can eventually add a timeout to disconnect if the client is not responding at all

internal/events/store.go
sashankg: yeah it is but can't hurt

internal/events/store.go
sashankg: no these need to be sent as individual messages with their own seq in order for the client to be able to reliably replay.

internal/events/store.go
sashankg: i'm following the firehouse api. i think it is possible for multiple records to be deleted at once. also can't hurt to keep the option available

internal/events/store.go
sashankg: yeah this isn't fully fleshed out yet. just wanted to demonstrate that the event object covers all event types

internal/events/store.go
sashankg: same this will eventually be part of the membership event type but that hasn't been fleshed out yet. it's not part of a space event

internal/events/store.go
sashankg: yeah might revisit this. the event object is different depending on the event type so we would need columns that were empty for some types. might be fine though

arushibandi (approved):

PR #384, opened 2026-06-07
[Spaces] Test UI for creating spaces

sashankg:

### DIFF
--- a/frontend/src/routes/_requireAuth/spaces/$space.tsx
+++ b/frontend/src/routes/_requireAuth/spaces/$space.tsx
@@ -113,14 +113,19 @@ function SpaceRecords() {
       <p className="text-sm text-muted-foreground mb-4">
         {records.length} record{records.length !== 1 ? "s" : ""}
       </p>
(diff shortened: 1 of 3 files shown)


arushibandi (approved):

frontend/src/routes/_requireAuth/spaces/$space.tsx
```diff
@@ -110,10 +109,25 @@ function SpaceRecords() {
...
       <p className="text-sm text-muted-foreground mb-4">
         {records.length} record{records.length !== 1 ? "s" : ""}
       </p>
+      <Button
+        onClick={async () => {
+          await procedure(
+            "network.habitat.space.putRecord",
+            {
+              space,
+              collection: "test.record.collection",
```
arushibandi, line 122: i've been wondering if we can create a syntax for internal-only / unvalidated lexicons. like `_test.record.collection` or something. So that organizations can define their own lexicons that don't need to be publicly published but can still be validated against

arushibandi (approved):

PR #386, opened 2026-06-08
[OAuthServer] Pass context to metrics

sashankg:


arushibandi (approved):
i honestly haven't been looking at these at all and the http traces should probably cover them? do we even need these metrics?

sashankg:
i think we need to consolidate them into just success vs failure but it is a good metric for overall health. plus we would want to setup alerts if there are too many auth failures

PR #387, opened 2026-06-09
[logins]: attempted fix

arushibandi:
I think we're sometimes overwriting the refresh token.

### DIFF
--- a/typescript/internal/src/authManager.ts
+++ b/typescript/internal/src/authManager.ts
@@ -148,7 +148,7 @@ export class AuthManager {
               this.setAuthState(token);
             }
           })
-          .then(() => {})
+          .then(() => { })
           .finally(() => {
             this.refreshPromise = undefined;
           });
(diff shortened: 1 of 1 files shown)


typescript/internal/src/authManager.ts
```diff
@@ -193,7 +193,7 @@ export class AuthManager {
     const state = {
       did: decoded.sub,
       accessToken: token.access_token,
-      refreshToken: token.refresh_token,
+      refreshToken: token.refresh_token ?? this.store.getState().authInfo?.refreshToken,
```
sashankg, line 196: pretty sure fosite always updates the refresh token on refreshes: https://github.com/ory/fosite/blob/a5f0b09bf31c17297b25637bb3fec2ff7a55b159/handler/oauth2/flow_refresh.go#L128

can we debug and see for sure if there's a way for it to be missing

arushibandi:
i'm pretty sure i saw network responses without it but i haven't had my
console open in a while. also this change wouldn't hurt right?

PR #390, opened 2026-06-10
[Sap] Init with auth

sashankg:
Setup sap binary that lets callers auth with pear. Will implement actual syncing next


cmd/sap/moon.yml
```diff
@@ -0,0 +1,7 @@
+tasks:
+  dev:
+    deps:
+      - root:caddy
+      - pear:dev
+    env:
+      SAP_SECRET: 'z42tt1ZWxkfKn5ujwLsELfY7191h4q6UCFjeRGf6tKXaMCnX'
```
arushibandi, line 7: did u mean to commit this?

frontend/moon.yml
```diff
@@ -10,9 +10,7 @@ tasks:
       HABITAT_DOMAIN: pear.local.habitat.network
       DOMAIN: frontend.local.habitat.network
     deps:
-      - target: pear:dev
-        env:
```
arushibandi, line 14: why?

cmd/pear/moon.yml
```diff
@@ -13,8 +13,9 @@ tasks:
       HABITAT_SERVICE_NAME: "habitat_local"
       HABITAT_DOMAIN: pear.local.habitat.network
       HABITAT_PRETTY_LOGS: 'true'
+      HABITAT_FRONTEND_DOMAIN: frontend.local.habitat.network
     deps:
-    - postgres
+    # - postgres
```
arushibandi: we can just remove it and use sqlite in dev

cmd/sap/flags.go
```diff
@@ -0,0 +1,46 @@
...
+			Name:    fLogLevel,
+			Usage:   "Log level (debug, info, warn, error)",
+			Value:   "info",
+			Sources: cli.EnvVars("SAP_LOG_LEVEL"),
+		},
+		&cli.StringFlag{
+			Name:    fSecret,
+			Usage:   "Secret used in OAuth flow",
+			Value:   "secret",
+			Sources: cli.EnvVars("SAP_SECRET"),
```
arushibandi, line 43: couldn't this be generated at init time? or does it need to be persistent

internal/sap/org_manager.go
```diff
@@ -0,0 +1,196 @@
...
+
+	"github.com/bluesky-social/indigo/atproto/atcrypto"
+	"github.com/bluesky-social/indigo/atproto/auth/oauth"
+	"github.com/bluesky-social/indigo/atproto/identity"
+	"github.com/bluesky-social/indigo/atproto/syntax"
+	"golang.org/x/oauth2"
+	"gorm.io/gorm"
+)
+
+type orgCredential struct {
```
arushibandi, line 18: does this need `gorm.Model`?

internal/sap/org_manager.go
```diff
@@ -0,0 +1,196 @@
...
+}
+
+var ErrOrgNotFound = fmt.Errorf("no org")
+
+func (o *orgManager) GetOrgs(ctx context.Context) ([]syntax.DID, error) {
+	var creds []orgCredential
+	err := o.db.WithContext(ctx).
+		Where("access_token != ''").
+		Where("expires_at > ?", time.Now()).
+		Find(&creds).
```
arushibandi, line 124: i think you could select the org DID column directly in this query?

cmd/sap/server.go
```diff
@@ -0,0 +1,88 @@
...
+	}
+	return s
+}
+
+func (s *server) handleHealth(w http.ResponseWriter, r *http.Request) {
+	w.Header().Set("Content-Type", "application/json")
+	_ = json.NewEncoder(w).Encode(map[string]string{"status": "ok"})
+}
+
+func (s *server) handleAddOrg(w http.ResponseWriter, r *http.Request) {
```
arushibandi, line 26: this is the route that's called when an appview / product wants an admin to sign up and add their org right?

arushibandi (approved):

cmd/sap/server.go
sashankg: yup

cmd/sap/moon.yml
sashankg: yeah i think it's fine to just have a hard coded secret for dev

frontend/moon.yml
sashankg: if there are different env between pear dependencies then we can't run `moon r sap:dev frontend:dev` because it will run 2 pears. I'll figure out a better dependency model later

cmd/sap/flags.go
sashankg: needs to be persistent

internal/sap/org_manager.go
sashankg: yeah will polish later

PR #393, opened 2026-06-11
[Sap] Basic subscriber

sashankg:
-------------------------
- master
  - **[Sap] Basic subscriber** :point_left:
    - #397


internal/sap/models.go
```diff
@@ -0,0 +1,27 @@
...
+type managedOrg struct {
+	DID  syntax.DID `gorm:"column:did;primaryKey"`
+	Host string
+
+	// Pending auth
+	State        *string `gorm:"index"`
+	CodeVerifier *string
+
+	// Completed auth
+	AccessToken  string `gorm:"type:text"`
```
arushibandi: why type text? is `string` not fine?

internal/sap/models.go
```diff
@@ -0,0 +1,27 @@
...
+	"github.com/bluesky-social/indigo/atproto/syntax"
+)
+
+type managedOrg struct {
+	DID  syntax.DID `gorm:"column:did;primaryKey"`
+	Host string
+
+	// Pending auth
+	State        *string `gorm:"index"`
+	CodeVerifier *string
```
arushibandi, line 15: why are these pointers?

internal/sap/models.go
```diff
@@ -0,0 +1,27 @@
...
+
+	// Completed auth
+	AccessToken  string `gorm:"type:text"`
+	RefreshToken string `gorm:"type:text"`
+	ExpiresAt    time.Time
+	CreatedAt    time.Time
+	UpdatedAt    time.Time
+
+	ErrorMsg string
```
arushibandi, line 24: ?

internal/sap/sap.go
```diff
@@ -21,6 +26,8 @@ type sapImpl struct {
 	*orgManager
 	db         *gorm.DB
 	pathPrefix string
+	subber     *subscriber
```
arushibandi: nit: can u name this `sub` `subber` is weird

internal/sap/sap.go
```diff
@@ -21,6 +26,8 @@ type sapImpl struct {
...
+	sseChan    chan *sse.Event
```
arushibandi: nit: usually i call them `sseCh` instead of `Chan`

internal/sap/subscriber.go
```diff
@@ -0,0 +1,97 @@
...
+}
+
+type subscriber struct {
+	db         *gorm.DB
+	orgManager *orgManager
+
+	mu            sync.RWMutex
+	subscriptions map[syntax.DID]*subscription
+
+	sseChan chan *sse.Event
```
arushibandi: nit: `sseCh`

internal/sap/subscriber.go
```diff
@@ -0,0 +1,97 @@
...
+
+type subscription struct {
+	client *sse.Client
+}
+
+type subscriber struct {
+	db         *gorm.DB
+	orgManager *orgManager
+
+	mu            sync.RWMutex
```
arushibandi: nit: add comment about what this is protecting 
```
mu            sync.RWMutex             // protects below
```

internal/sap/subscriber.go
```diff
@@ -0,0 +1,97 @@
...
+		sseChan:       sseChan,
+	}
+}
+
+func (s *subscriber) addSubscription(ctx context.Context, org *managedOrg) {
+	client := sse.NewClient(org.Host + "/xrpc/network.habitat.sync.subscribeSpaces")
+	client.Connection = s.orgManager.GetClient(ctx, org.DID)
+	client.LastEventID.Store([]byte(org.Cursor))
+	sub := &subscription{client: client}
+	// client.ReconnectStrategy = &backoff.StopBackOff{}
```
arushibandi: why commented out?

internal/sap/subscriber.go
```diff
@@ -0,0 +1,97 @@
...
+	err := client.SubscribeChanRawWithContext(context.Background(), s.sseChan)
```
arushibandi: why not `ctx`?

internal/sap/subscriber.go
```diff
@@ -0,0 +1,97 @@
...
+	"errors"
+	"fmt"
+	"log/slog"
+	"sync"
+
+	"github.com/bluesky-social/indigo/atproto/syntax"
+	"github.com/r3labs/sse/v2"
+	"gorm.io/gorm"
+)
+
```
arushibandi, line 14: comments in this file ?

internal/sap/models.go
sashankg: they can be null

internal/sap/models.go
sashankg: if there's an error message when trying to connect to them

internal/sap/subscriber.go
sashankg: the sse client has a default exponential backoff. forgot to remove the comment

internal/sap/subscriber.go
sashankg: the subscription has a longer lifetime than this function

internal/sap/models.go
arushibandi: feels like this doesn't need to be stored but ok

arushibandi (approved):

internal/sap/models.go
sashankg: orgs will be connected to after they are persisted. if they are persisted successfully, that means we have valid auth and should be able to connect to them. I don't think an error means we need to evict the org from storage. instead, we just need to inform the client about the error when they check on status. we can attempt connecting again later.

PR #397, opened 2026-06-12
[Sap] Outbox

sashankg:
Will figure out how to actual deliver messages from the outbox later but sap will just store messages for now


internal/events/store.go
```diff
@@ -48,7 +48,12 @@ type EventOps struct {
 	Access string
 }
 
+type Subscriber interface {
+	Subscribe(ctx context.Context, since uint64) <-chan Event
```
arushibandi, line 53: naming is a bit weird here ... the subscriber is the thing calling `something.Subscribe(..)` and the thing implementing `Subscribe(..)` is `EventStream` or something similar

internal/events/store.go
```diff
@@ -48,7 +48,12 @@ type EventOps struct {
...
+}
+
 type Store interface {
```
arushibandi: would love comments on all important types

internal/sap/models.go
```diff
@@ -25,3 +28,30 @@ type managedOrg struct {
...
+	RepoStateActive   repoState = "active"
+	RepoStateDesynced repoState = "desynced"
+)
+
+type managedRepo struct {
+	Space habitat_syntax.SpaceURI `gorm:"primaryKey"`
+	DID   syntax.DID              `gorm:"column:did;primaryKey"`
+	Rev   syntax.TID
+	State repoState `gorm:"index"`
+}
```
arushibandi, line 47: would love comments on types

internal/sap/org_manager.go
```diff
@@ -22,12 +22,8 @@ type orgManager struct {
...
 }
 
-func newOrgManager(db *gorm.DB, domain string, secret atcrypto.PrivateKey) (*orgManager, error) {
-	err := db.AutoMigrate(&managedOrg{})
-	if err != nil {
-		return nil, err
-	}
-	return &orgManager{db: db, domain: domain, secret: secret}, nil
+func newOrgManager(db *gorm.DB, domain string, secret atcrypto.PrivateKey) *orgManager {
+	return &orgManager{db: db, domain: domain, secret: secret}
```
arushibandi: why did the AutoMigrate get removed?

internal/sap/sap.go
```diff
@@ -37,24 +34,21 @@ type SapConfig struct {
 }
 
 func NewSap(config SapConfig) (Sap, error) {
+	if err := autoMigrate(config.DB); err != nil {
```
arushibandi, line 38: what is `autoMigrate`?

internal/sap/sap.go
```diff
@@ -37,24 +34,21 @@ type SapConfig struct {
 }
```
arushibandi, line 35: just take these in as constructor arguments rather than a `config`. it's harder to debug and understand what the values in `config` is

internal/sap/subscriber.go
```diff
@@ -2,18 +2,21 @@ package sap
...
 
 	"github.com/bluesky-social/indigo/atproto/syntax"
+	"github.com/habitat-network/habitat/internal/events"
 	"github.com/r3labs/sse/v2"
 	"gorm.io/gorm"
 )
 
 type subscription struct {
 	client *sse.Client
+	cancel context.CancelFunc
```
arushibandi, line 19: usually `cancel` functions should be returned to the caller / via the constructor. so the caller does lifetime management.

arushibandi: i think what you can do is have one context per `subscriber`, and then have each subscription's context be a child of that. so that in the `closeSubscriptions` fn you can just cancel that one context.

internal/sap/subscriber.go
```diff
@@ -40,17 +40,108 @@ func (s *subscriber) addSubscription(ctx context.Context, org *managedOrg) {
...
+			if err != nil {
+				slog.ErrorContext(subscribeCtx, "failed to save space event", "err", err)
+				return
+			}
+		default:
+			slog.WarnContext(subscribeCtx, "unknown event type", "event", event.Event)
+		}
+	})
+	if err != nil {
+		slog.ErrorContext(ctx, "failed to subscribe", "org", org.DID, "err", err)
```
arushibandi, line 134: would need to cancel the subscriptions context here as well (if not cancelled, the context hangs around in memory)
probably worth having a cleanup function for subscriptions since you do it twice which cancels the context + deletes by did.

internal/sync/server.go
```diff
@@ -12,10 +12,10 @@ import (
 )
 
 type Server struct {
-	eventStore events.Store
+	eventStore events.Subscriber
```
arushibandi: ya i think maybe this is called stream not susbcriber?>=

internal/sap/org_manager.go
sashankg: moved all the model migrations to a single spot. the components in sap just use gorm as the store abstraction. the various managers operate directly on multiple gorm models

internal/sap/sap.go
sashankg: single function to migrate all models

internal/sap/sap.go
sashankg: this is going to be exposed to library users and there will be lots of config options for tuning the performance of sync operations. Will leave as config object for now since this will keep changing

internal/sap/subscriber.go
sashankg: we are still going to want to be able to close individual subscribers by org id so i think we still need to track the cancel function per subscriber. 

> usually cancel functions should be returned to the caller

the caller here is either `loadSubscriptions` which checks the persisted orgs and boots up subcriptions for each one or the `/org/add` handler. neither of them i think should be responsible for managing the lifetime. I think the "subscriber" should be responsible for managing all subscriptions

internal/sap/subscriber.go
sashankg: moved adding the subscriber to `subscriptions` to after the error check so it won't be persisted in memory in the error case

arushibandi (approved):

PR #398, opened 2026-06-16
[Sap] Crawler

sashankg:


internal/sap/crawler.go
```diff
@@ -0,0 +1,180 @@
...
+	"net/http"
+	"net/url"
+
+	"github.com/bluesky-social/indigo/atproto/syntax"
+	"github.com/habitat-network/habitat/api/habitat"
+	habitat_syntax "github.com/habitat-network/habitat/internal/syntax"
+	"gorm.io/gorm"
+	"gorm.io/gorm/clause"
+)
+
```
arushibandi, line 17: comments pls

internal/sap/crawler.go
```diff
@@ -0,0 +1,180 @@
...
+	}
+	return nil
+}
+
+func (c *crawler) crawlOrg(ctx context.Context, org *managedOrg) {
+	if err := c.db.WithContext(ctx).
+		Model(&managedOrg{}).
+		Where("did = ?", org.DID).
+		Update("crawl_state", crawlStateRunning).Error; err != nil {
+		slog.ErrorContext(ctx, "set crawl running", "org", org.DID, "err", err)
```
arushibandi, line 52: does this ever get retried? should we set a todo? alternatively you could have `crawlOrg` return (fn, err) and if err == nil then do go fn()

internal/sap/crawler.go
```diff
@@ -0,0 +1,180 @@
...
+
+	if crawlErr != nil {
+		if err := c.db.WithContext(ctx).
+			Model(&managedOrg{}).
+			Where("did = ?", org.DID).
+			Updates(map[string]any{
+				"crawl_state": crawlStateErrored,
+				"error_msg":   crawlErr.Error(),
+			}).Error; err != nil {
+			slog.ErrorContext(ctx, "set crawl errored", "org", org.DID, "err", err)
```
arushibandi, line 66: doesn't this mean it is a broken state ? like the crawl_state could be something other than errored then it never gets retried?

internal/sap/crawler.go
```diff
@@ -0,0 +1,180 @@
...
+		}
+		// TODO: cancel subscriptions
+		slog.ErrorContext(ctx, "crawl failed", "org", org.DID, "err", crawlErr)
+		return
+	}
+
+	if err := c.db.WithContext(ctx).
+		Model(&managedOrg{}).
+		Where("did = ?", org.DID).
+		Update("crawl_state", crawlStateComplete).Error; err != nil {
```
arushibandi, line 76: wait i'm confused how did it complete? shouldn't it always be waiting on new updates even if its done

internal/sap/crawler.go
```diff
@@ -0,0 +1,180 @@
...
+
+		for _, space := range listSpacesOutput.Spaces {
+			if err := c.enumerateSpaceMembers(ctx, client, org, space.Uri); err != nil {
+				slog.ErrorContext(ctx, "enumerate space members", "space", space.Uri, "err", err)
+			}
+		}
+
+		if listSpacesOutput.Cursor == "" {
+			break
+		}
```
arushibandi, line 131: whY?

internal/sap/crawler.go
```diff
@@ -0,0 +1,180 @@
...
+	if resp.StatusCode != http.StatusOK {
+		return fmt.Errorf("get members: %s", resp.Status)
+	}
+
+	space := habitat_syntax.SpaceURI(spaceURI)
+	for _, member := range getMembersOutput.Members {
+		if err := c.db.WithContext(ctx).
+			Clauses(clause.OnConflict{DoNothing: true}).
+			Create(&managedRepo{
+				Space: space,
```
arushibandi, line 172: nit: should this be `managedSpaceRepo`

arushibandi (approved):

internal/sap/crawler.go
sashankg: yeah added a todo

internal/sap/crawler.go
sashankg: yeah will figure it out when we actually add the retry mechanism later

internal/sap/crawler.go
sashankg: no new updates will come from `subscribeSpaces` crawl is just for backfillling

internal/sap/crawler.go
sashankg: if listSpaces doesn't return a cursor then there are no more spaces to list so we can stop

internal/sap/crawler.go
sashankg: i think it's implied

PR #401, opened 2026-06-16
[identity]: separate server from hive library and add auth for DIDs and handles

arushibandi:
as discussed

### DIFF
--- a/internal/identity/server.go
+++ b/internal/identity/server.go
@@ -142,7 +142,13 @@ func (s *Server) ServeDIDDoc(w http.ResponseWriter, r *http.Request) {
 	}
 	callerOrg, err := s.orgStore.GetOrgForDID(r.Context(), callerDID)
 	if err != nil {
-		utils.LogAndHTTPError(r.Context(), w, err, "looking up caller org", http.StatusInternalServerError)
+		utils.LogAndHTTPError(
+			r.Context(),
+			w,
+			err,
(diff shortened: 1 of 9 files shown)


internal/identity/server.go
```diff
@@ -150,12 +135,35 @@ func (s *Server) GetServiceAuth(w http.ResponseWriter, r *http.Request) {
 
 // Serve DID Doc ( satisfy /{did}/.well-known/did.json )
 func (s *Server) ServeDIDDoc(w http.ResponseWriter, r *http.Request) {
+	// Must present valid oauth credential for this org to read identities
```
sashankg, line 138: let's make an exception for the org's did doc

internal/identity/server.go
```diff
@@ -150,12 +135,35 @@ func (s *Server) GetServiceAuth(w http.ResponseWriter, r *http.Request) {
...
+	callerDID, ok := authn.Validate(w, r, s.oauth)
```
sashankg, line 139: this will break oauth server who just does a regular http request to fetch the did info:

https://github.com/habitat-network/habitat/blob/5d5c563be81d472c53d5ab5b5b40ab79012de926/internal/oauthserver/oauth_server.go#L300

we'll need to pass a custom implementation of directory that directly reads from hive

sashankg (approved):
make sure to just test it manually as well before merging

PR #403, opened 2026-06-16
[Sap] Resyncer

sashankg:
Final component of Sap. While crawling, the subscriber writes to a resync buffer that will get drained once crawl is complete. Any desynced or pending repos are picked up by resyncer workers that will use `getRepoOplog` to catch up to the current state.

-------------------------
- master
  - **[Sap] Resyncer** :point_left:
    - #407


internal/sap/crawler.go
```diff
@@ -16,17 +16,26 @@ import (
 )
```
arushibandi, line 17: comments

arushibandi, line 17: also would be nice to have a top-level `README.md` in `internal/sap` since this is bound to get complicated and confusing

internal/sap/resyncer.go
```diff
@@ -0,0 +1,303 @@
...
+
+func newResyncer(
+	db *gorm.DB,
+	orgManager *orgManager,
+	resyncBuf *resyncBuffer,
+	resyncNotifCh chan struct{},
+	parallelism int,
+) *resyncer {
+	if parallelism <= 0 {
+		parallelism = 5
```
arushibandi, line 43: i think you can look up gomaxprocs or something and just set it to the number of cpu cores the computer has

internal/sap/resyncer.go
```diff
@@ -0,0 +1,303 @@
...
+					CASE state
+						WHEN "pending" THEN 1
+						WHEN "desynced" THEN 2
+						WHEN "error" THEN 3
+					END,
+					rowid
+				LIMIT ?
+			)
+			RETURNING space, did
+		`, now, 100).Rows()
```
arushibandi, line 95: how did u pick this batch size

internal/sap/resyncer.go
```diff
@@ -0,0 +1,303 @@
...
+
+		if len(output.Records) > 0 {
+			err = r.db.WithContext(ctx).Transaction(func(tx *gorm.DB) error {
+				if err := writeOplogRecords(tx, space, repoDID, output.Records); err != nil {
+					return err
+				}
+				lastRev := syntax.TID(output.Records[len(output.Records)-1].Rev)
+				return tx.Model(&managedRepo{}).
+					Where("space = ? AND did = ?", space, repoDID).
+					Update("rev", lastRev).Error
```
arushibandi, line 251: do these DB writes need to happen atomically?

internal/sap/resyncer.go
```diff
@@ -0,0 +1,303 @@
...
+		Updates(map[string]any{
+			"state":       RepoStateError,
+			"error_msg":   errMsg,
+			"retry_count": retryCount,
+			"retry_after": retryAfter.Unix(),
+		}).Error
+}
+
+func backoff(retries int, maxMinutes int) time.Duration {
+	dur := 1 << retries
```
arushibandi, line 324: use this package: https://pkg.go.dev/github.com/bradenaw/retry

internal/sap/resync_buffer.go
```diff
@@ -0,0 +1,221 @@
...
+	"fmt"
+	"log/slog"
+
+	"github.com/bluesky-social/indigo/atproto/syntax"
+	"github.com/habitat-network/habitat/internal/db"
+	"github.com/habitat-network/habitat/internal/events"
+	"gorm.io/gorm"
+)
+
+type resyncBuffer struct {
```
arushibandi, line 19: comments ?

internal/sap/resync_buffer.go
```diff
@@ -0,0 +1,221 @@
...
+		resyncNotifCh: rb.resyncNotifCh,
+	}
+}
+
+func (rb *resyncBuffer) shouldBuffer(org *managedOrg, repo *managedRepo) bool {
+	if org.CrawlState != nil && *org.CrawlState == crawlStateRunning {
+		return true
+	}
+	if repo == nil {
+		return true
```
arushibandi, line 42: what does a nil repo mean?

internal/sap/resyncer.go
```diff
@@ -0,0 +1,303 @@
...
+	"net/http"
+	"net/url"
+	"time"
+
+	"github.com/bluesky-social/indigo/atproto/syntax"
+	"github.com/habitat-network/habitat/api/habitat"
+	habitat_syntax "github.com/habitat-network/habitat/internal/syntax"
+	"gorm.io/gorm"
+)
+
```
arushibandi, line 19: comments

internal/sap/sap.go
```diff
@@ -74,3 +90,8 @@ func (s *sapImpl) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 	srv := &server{sap: s}
 	srv.ServeHTTP(w, r)
 }
+
+func (s *sapImpl) startOrgSync(org *managedOrg) {
+	go s.crawler.crawlOrg(context.Background(), org)
+	go s.sub.addSubscription(context.Background(), org)
```
arushibandi, line 96: feels like there could be a race condition here

arushibandi (commented):
something kinda feels like this could be simplified a bit ... but maybe i just need to wrap my head around it more

internal/sap/resyncer.go
sashankg: arbitrarily. will make it configurable later

internal/sap/resyncer.go
sashankg: we need the outbox and the managedRepo rev to be in sync so we update those together

internal/sap/resyncer.go
sashankg: that library is not useful here. we aren't trying to retry a specific function. we are persisting the retry state in the db to be read later.

internal/sap/resync_buffer.go
sashankg: the crawler hasn't discovered this repo yet. if the event has a "since" that means there are previous repo events we need to backfill. otherwise it's a new repo and we don't need to buffer. fixed

internal/sap/sap.go
sashankg: addSubscription passes its events to the resync buffer. the resync buffer determines if it should write to the outbox directly or buffer the event until the crawl + resync is complete. that's how we avoid the race

internal/sap/resyncer.go
arushibandi: oh i meant the writeOpLogRecords and the lastRev write

internal/sap/resyncer.go
```diff
@@ -0,0 +1,330 @@
...
+	retryAfter := time.Now().Add(backoff(retryCount, 60))
+	errMsg := ""
+	if syncErr != nil {
+		errMsg = syncErr.Error()
+	}
+	return r.db.WithContext(ctx).
+		Model(&managedRepo{}).
+		Where("space = ? AND did = ?", space, did).
+		Updates(map[string]any{
+			"state":       RepoStateError,
```
arushibandi, line 316: when does this actually get retried?

internal/sap/outbox.go
```diff
@@ -22,3 +26,33 @@ func writeEventOps(tx *gorm.DB, ops []events.EventOps) error {
 	}
 	return nil
 }
+
+func writeOplogRecords(
```
arushibandi, line 30: i'm confused who consumes from the outbox?

internal/sap/resyncer.go
```diff
@@ -0,0 +1,330 @@
...
+					"repo",
+					job.DID,
+					"err",
+					err,
+				)
+			}
+		}
+	}
+}
+
```
arushibandi, line 154: still think we can use more comments e.g, in this file

arushibandi (commented):
did a readme get added ? would prefer to do this now and we can update as we go

arushibandi (approved):
approving but would like to see remaining comments addresesed

PR #407, opened 2026-06-17
[Sap] Integration test

sashankg:


cmd/sap/sap_test.go
```diff
@@ -0,0 +1,311 @@
...
+		pearServer.Close()
+	})
+	go func() {
+		if err := s.Start(ctx); err != nil {
+			t.Logf("Start returned: %v", err)
+		}
+	}()
+
+	// 5. Create remaining data before the crawl runs.
+	// This ensures all spaces and records exist when the crawl lists spaces,
```
arushibandi: we probably should also test correctness when there is both a crawl + live data creation happening

cmd/sap/sap_test.go
```diff
@@ -0,0 +1,311 @@
...
+		Value []byte
+	}
+	var outboxRecords []localOutbox
+	err = db.Table("outboxes").Find(&outboxRecords).Error
+	require.NoError(t, err)
+
+	for _, rec := range outboxRecords {
+		t.Logf("Outbox Record: ID=%d, URI=%s", rec.ID, rec.URI)
+		require.Contains(t, createdURIs, rec.URI)
+	}
```
arushibandi, line 217: also need to check that createdURIs is empty after seeing all the outboxRecords right?

arushibandi (approved):

cmd/sap/sap_test.go
```diff
@@ -0,0 +1,311 @@
...
+			backfillSpaceURI,
+			adminId.DID,
+			collection,
+			rkey,
+			map[string]any{"data": fmt.Sprintf("live-update-%d", j)},
+		)
+		require.NoError(t, err)
+		createdURIs[recURI.String()] = true
+	}
+
```
arushibandi, line 189: confused about this

PR #409, opened 2026-06-18
[ods]: instance admin to set config for the entire instance

arushibandi:
This PR just sets up an admin + password and an empty page that will have config settings. This page is just hosted by pear, so in local it is at `pear.local.habitat.network/admin`. That makes it slightly simpler for deployment purposes, now there's just one binary in the docker image.

Should the login/logout endpoints be XRPCs? I didn't see that for the `com.atproto.admin` APIs


internal/instanceadmin/models.go
```diff
@@ -0,0 +1,22 @@
+package instanceadmin
+
+import "time"
+
+// instanceAdminAccountID is the fixed primary key for the single instance admin row.
+const instanceAdminAccountID = 1
```
arushibandi: is this correct / typical?

internal/instanceadmin/store.go
```diff
@@ -0,0 +1,156 @@
...
+		return false, err
+	}
+	ok, err := argon2id.ComparePasswordAndHash(password, account.PasswordHash)
+	if errors.Is(err, argon2id.ErrInvalidHash) {
+		return false, nil
+	}
+	return ok, err
+}
+
+func (s *storeImpl) CreateSession(ctx context.Context) (string, time.Time, error) {
```
sashankg: use https://github.com/gorilla/sessions

cmd/pear/main.go
```diff
@@ -128,6 +129,28 @@ func run(_ context.Context, cmd *cli.Command) error {
...
+		startupCtx,
+		cmd.String(fAdminPassword),
+	)
+	if err != nil {
+		slog.Error("unable to bootstrap instance admin", "err", err)
+		os.Exit(1)
+	}
+	if adminCreated && adminPassword != "" {
+		slog.Warn(
+			"generated instance admin password; save it now, it will not be shown again",
```
sashankg: let's just keep this simple for now and not persist the password. either pass it in at start time or a random will be generated every startup.

internal/instanceadmin/home.html
sashankg, line 1: we should create a new typescript package for this UI and point the build dir to be embedded by pear. during dev we can setup a reverse proxy to point to the dev server instead.

sashankg (approved):

internal/instanceadmin/home.html
arushibandi: talked irl, this does not include the org flow and is just the admin flow so we can leave raw html here

PR #410, opened 2026-06-18
[ods]: support admin configuration and invite links for orgs to join

arushibandi:
Screenshot 2026-06-18 at 9 13 39 AM
Screenshot 2026-06-18 at 9 13 46 AM


cmd/pear/main.go
```diff
@@ -643,3 +648,29 @@ func corsMiddleware(next http.Handler) http.Handler {
...
+		pass = string(b)
+		slog.Warn(
+			"generated instance admin password; save it now, it will not be shown again until next restart. password changes on restart if not added to environment variables via HABITAT_ADMIN_PASSWORD",
+			"username",
+			"admin",
+			"password",
+			pass,
+		)
+	}
+	passwordHash, err := argon2id.CreateHash(pass, argon2id.DefaultParams)
```
sashankg, line 671: do we need to hash it if it's just in memory?

internal/instance/store.go
```diff
@@ -0,0 +1,344 @@
...
+	var settings instanceSettings
+	err := s.db.WithContext(ctx).First(&settings, instanceSettingsID).Error
+	if err == nil {
+		return &settings, nil
+	}
+	if !errors.Is(err, gorm.ErrRecordNotFound) {
+		return nil, err
+	}
+
+	secret, err := generateSigningSecret()
```
sashankg: can't we just use the same secret as the cookie store? why do we need another one

internal/instance/store.go
```diff
@@ -0,0 +1,344 @@
...
+	if err != nil {
+		return nil, err
+	}
+	settings = instanceSettings{
+		ID:                instanceSettingsID,
+		OrgCreationPolicy: policyOpen,
+		SigningSecret:     secret,
+		CreatedAt:         time.Now(),
```
sashankg: gorm does this automtically as long as the model as `CreatedAt`

internal/instance/store.go
```diff
@@ -0,0 +1,344 @@
...
+		return "", err
+	}
+	jti := hex.EncodeToString(b)
+	now := time.Now()
+	expiresAt := now.Add(inviteDuration)
+
+	invite := instanceInvite{
+		Token:     jti,
+		CreatedAt: now,
+		ExpiresAt: expiresAt,
```
sashankg: why do we need to persist the expiresat?

frontend/src/routes/org/create.tsx
```diff
@@ -40,6 +62,53 @@ function PasswordInput(props: React.ComponentProps<typeof Input>) {
 
 function CreateOrgPage() {
   const navigate = useNavigate();
+  const { token } = Route.useSearch();
+  const invite = token ? decodeInviteToken(token) : null;
+
+  const [useCustomInstance, setUseCustomInstance] = useState(false);
+  const [customDomain, setCustomDomain] = useState("");
+  const [customInstanceName, setCustomInstanceName] = useState<string | null>(
```
sashankg: this should be values in the form

sashankg (approved):

cmd/pear/main.go
arushibandi: i guess not? idk seems weird

internal/instance/store.go
arushibandi: sure that works

cmd/pear/main.go
arushibandi: i think you're still supposed to hash due to timing attacks ?

PR #413, opened 2026-06-19
[Sap] Delivery

sashankg:
Add functionality to listen to changes to the outbox, fetch them, and ack messages. This lets callers consume the outbox at their own pace. Added a sample websocket delivery system to the sap binary as a demo


internal/sap/outbox.go
```diff
@@ -56,3 +58,82 @@ func writeOplogRecords(
...
+}
+
+// notifyOutbox signals ch that new outbox rows may be available, without
+// blocking if a notification is already pending.
+func notifyOutbox(ch chan struct{}) {
+	select {
+	case ch <- struct{}{}:
+	default:
+	}
+}
```
arushibandi, line 139: i don't really get this. if the caller has ch already, can't it just do this send itself ?

cmd/sap/websocket.go
```diff
@@ -0,0 +1,102 @@
...
+			msg, ok := pending[id]
+			if !ok {
+				continue
+			}
+			if err := msg.Ack(ctx); err != nil {
+				slog.ErrorContext(ctx, "ack outbox message", "id", id, "err", err)
+				continue
+			}
+			delete(pending, id)
+		case <-s.sap.Watch():
```
arushibandi, line 99: i also don't get this -- shouldn't we be doing fan-out on every notify from the outbox? otherwise there could be a client that never wins the race here and just gets out of date?

internal/sap/outbox.go
```diff
@@ -56,3 +58,82 @@ func writeOplogRecords(
...
+	ID    uint
+	URI   habitat_syntax.SpaceRecordURI
+	Value json.RawMessage
+
+	ack func(ctx context.Context) error
+}
+
+// Ack marks the message as processed so it is not redelivered by Poll.
+func (m OutboxMessage) Ack(ctx context.Context) error {
+	return m.ack(ctx)
```
arushibandi, line 75: i'm confused how this works for all callers? seems li i feel like this should be part of `websocket.go` and `Poll()` should take in the last-acked on a per-caller basis. otherwise if one callers acks doesn't that get marked for all? or am i missing something?

internal/sap/outbox.go
sashankg: yeah it's just a helper for dropping the notification when it's already been notified. Added a simpler interface in https://github.com/habitat-network/habitat/pull/414

internal/sap/outbox.go
```diff
@@ -56,3 +58,82 @@ func writeOplogRecords(
...
+	// fan-out: only one consumer should be draining it at a time.
+	Watch() <-chan struct{}
+}
+
+type outboxImpl struct {
+	db       *gorm.DB
+	notifyCh chan struct{}
+}
+
+func newOutbox(db *gorm.DB, notifyCh chan struct{}) *outboxImpl {
```
arushibandi, line 94: this pattern is weird where the caller gives a channel and then the caller calls `Watch()` to get the same channel back

NewOutbox should return a channel to listen on or remove Watch()

sashankg: yeah the caller of `newOutbox` isn't expcted to use watch. `Sap` embeds the outbox so its callers can use watch. I guess `Sap` can be responsible for implementing Outbox itself. I'll reorganize this in a follow up

internal/sap/outbox.go
arushibandi: ok will wait for that to merge and then rebase on that

internal/sap/outbox.go
arushibandi: might as well do it in this one, no?

arushibandi (approved):

PR #418, opened 2026-06-19
[OAuthServer] Support JWT bearer credentials

sashankg:
We can allow some clients to just use their private key to receive a token instead of requiring an admin to authorize it. This will let approved services like the admin and search service not require an admin to manually auth


internal/oauthserver/fosite_storage.go
```diff
@@ -71,6 +76,20 @@ func (s *store) ClientAssertionJWTValid(ctx context.Context, jti string) error {
...
+	return &client{*metadata}, nil
+}
+
+// fetchClientMetadata fetches and decodes the client metadata document
+// published at id (the client's client_id URL). See
+// https://atproto.com/specs/oauth#client-id-metadata-document.
+func (s *store) fetchClientMetadata(
+	ctx context.Context,
+	id string,
+) (*pdsclient.ClientMetadata, error) {
```
arushibandi, line 101: why not just return `pdsclient.ClientMetadata` so you don't have to do `*metadata` in `GetClient`?

internal/oauthserver/jwt_bearer.go
```diff
@@ -0,0 +1,21 @@
...
+// use the JWT Bearer grant (RFC 7523, urn:ietf:params:oauth:grant-type:jwt-bearer)
+// to mint access tokens directly from a signed assertion, without a
+// user-driven authorization flow. The assertion's "iss" claim must match one
+// of these client IDs, and is verified against the JWKS published in that
+// client's metadata document.
+//
+// TODO: replace with a persisted/dynamic registry once client registration
+// for this grant type is supported.
+var jwtBearerAllowedClients = map[string]struct{}{
+	// "https://example.com/client-metadata.json": {},
```
arushibandi: this may have to be derived from some env vars i think

internal/oauthserver/oauth_server.go
```diff
@@ -197,19 +200,25 @@ func NewOAuthServer(
...
 	meter metric.Meter,
 	orgStore org.Store,
+	domain string,
 ) (*OAuthServer, error) {
 	config := &fosite.Config{
 		GlobalSecret:               secret,
 		SendDebugMessagesToClients: true,
 		RefreshTokenScopes:         []string{},
 		ScopeStrategy:              scopeStrategy,
+		TokenURL:                   "https://" + domain + "/oauth/token",
```
arushibandi: i'm confused why do we need this? isn't this its own endpoint?

internal/oauthserver/jwt_bearer_test.go
```diff
@@ -0,0 +1,256 @@
...
+		otherKey, err := ecdsa.GenerateKey(elliptic.P256(), rand.Reader)
+		require.NoError(t, err)
+
+		resp := postJWTBearerToken(
+			t,
+			tokenURL,
+			client.assertion(t, "did:web:subject.example", tokenURL, "jti-3", otherKey),
+		)
+		defer func() { _ = resp.Body.Close() }()
+		require.NotEqual(t, http.StatusOK, resp.StatusCode)
```
arushibandi: maybe check specific status codes instead of not equal (401 unauth?)

internal/oauthserver/fosite_storage.go
```diff
@@ -92,7 +111,89 @@ func (s *store) GetClient(ctx context.Context, id string) (fosite.Client, error)
...
+	}
+	return cl.GetScopes(), nil
+}
+
+func (s *store) isJWTBearerClientAllowed(clientID string) bool {
+	metadataURL, err := url.Parse(clientID)
+	if err != nil {
+		return false
+	}
+	// support any subdomains for now
```
arushibandi: why? we can just only support the hardcoded ones to be safe right?

arushibandi (approved):

internal/oauthserver/oauth_server.go
sashankg: the rfc7523 handler validates that the audience of the jwt is the expected token endpoint

internal/oauthserver/fosite_storage.go
sashankg: rn i'm only passing in the domain of pear itself so keeping it as subdomains of pear. will update later

PR #419, opened 2026-06-19
[login UI]: allow different hosting providers

arushibandi:


typescript/internal/src/components/HostingSelector.tsx
```diff
@@ -0,0 +1,97 @@
...
+    value !== defaultDomain ? value : "",
+  );
+  const [customInstanceName, setCustomInstanceName] = useState<string | null>(
+    null,
+  );
+  const [customInstanceError, setCustomInstanceError] = useState<
+    string | null
+  >(null);
+
+  useEffect(() => {
```
sashankg, line 33: there should not be a useEffect. this will call `describeInstance` on every key stroke. can we not just do it on submit?

typescript/internal/src/components/HostingSelector.tsx
sashankg, line 1: this file should be in frontend/. it won't be used by other apps, right?

typescript/internal/src/authManager.ts
```diff
@@ -5,39 +5,35 @@ import { create, StoreApi } from "zustand";
...
   };
-  private config: client.Configuration;
   private onUnauthenticated: () => void;
   private refreshPromise: Promise<void> | undefined;
 
   constructor(
     appName: string,
     domain: string,
-    serverDomain: string,
+    defaultDomain: string,
```
sashankg, line 31: we should be able to remove this entirely, right? the caller of loginUrl can provide a default serverdomain if they want

sashankg (approved):

arushibandi:
Closing this in favor of keeping DIDs / handles public for now with a plan to support all of 1) no handles 2) private handles 3) public handles in the future, configurable by org.

PR #420, opened 2026-06-20
Add OAuth consent flow for org DID authorization

sashankg:
## Summary
Implements an explicit consent screen for organization DID OAuth flows. When an admin authenticates on behalf of an org, they must now review and approve the client's access request before an authorization code is issued. Individual member DID flows bypass this step.

## Key Changes
- **Consent page handler** (`HandleConsent`, `HandleConsentDecision`): Renders client metadata and requested scopes; processes admin's allow/deny decision
- **Consent session management**: Stores pending authorization state in signed, encrypted cookies with 10-minute expiration and path-scoped invalidation
- **Org DID detection**: `HandleAuthorize` now checks if the authenticated DID belongs to an org; if so, sets `RequiresConsent` flag
- **Callback routing**: `HandleCallback` redirects org DID flows to consent page instead of immediately issuing a code
- **Flash cookie refactoring**: Migrated from `gob`-encoded flashes to explicit key-value storage in session; added `RequiresConsent` field to `authRequestFlash`
- **Helper methods**: Added `popFlash()` and `recreateAuthorizeRequest()` to reduce duplication between callback and consent handlers
- **Consent UI**: Added `consent.html` template displaying client name/logo, org DID, and requested scopes with allow/deny buttons
- **Comprehensive tests**: `TestOAuthServerOrgDIDRequiresConsent` validates consent page rendering, allow decision issuance, and deny rejection

## Implementation Details
- Consent and flash cookies use the same encryption key (already reused elsewhere) and are scoped to their respective paths (`/oauth/consent` and `/oauth-callback`) to prevent cross-endpoint replay
- Session options are explicitly set at creation time and re-applied during invalidation to ensure Set-Cookie path matching
- Authorization request form data is preserved across redirects by URL-encoding it in the session rather than relying on server-side state

https://claude.ai/code/session_01EqSyq7aMnPYirqaghmWc9X


internal/oauthserver/consent.go
```diff
@@ -0,0 +1,240 @@
...
+
+var consentTemplates = template.Must(template.ParseFS(consentTemplateFS, "consent.html"))
+
+// consentPageData is rendered into consent.html.
+type consentPageData struct {
+	ClientName string
+	ClientUri  string
+	LogoUri    string
+	Scopes     []string
+	OrgDID     string
```
arushibandi: nit: can we  use handle instead of DID here?

internal/oauthserver/consent.go
```diff
@@ -0,0 +1,240 @@
...
+	if !ok {
+		return nil, http.ErrNoCookie
+	}
+	form, err := url.ParseQuery(formStr)
+	if err != nil {
+		return nil, err
+	}
+	did, _ := session.Values[consentDidKey].(string)
+
+	if pop {
```
arushibandi: this pattern makes the code harder to read and is generally bad where the caller needs to know what this arg is gonna do in the function. can we refactor the functions to not have to pass this in and just have the caller do this if needed ? since the caller will have access to o.sessionStore as well

arushibandi (approved):

PR #422, opened 2026-06-22
[Spaces] Support empty member for listSpaces

sashankg:


internal/spaces/spaces.go
```diff
@@ -237,27 +230,31 @@ func (s *store) ListSpaces(
...
-	conditions := s.db.Where("owner = ?", member)
-	for _, key := range fgaObjects {
-		uri, err := fgastore.ParseSpaceObjectKey(key)
+	var conditions *gorm.DB
+	if member == "" {
+		conditions = s.db
+	} else {
+		// If member is the org itself, include all the org's spaces. This initial condition
+		// is always false for user members since they are never owners of a space.
+		conditions = s.db.Where("owner = ?", member)
```
arushibandi, line 242: wait shouldn't this be the member's org then ? otherwise nothing would be found?

arushibandi (approved):

internal/spaces/spaces.go
sashankg: member could be the org itself that's what this condition is accounting for. for literal members of the org, we include other conditions based on what they have permission to below

PR #423, opened 2026-06-22
[search]: implement

arushibandi:
A basic search implementation that uses sap and indexes a single org's records, without any within-org permissions.


cmd/search/extract_test.go
```diff
@@ -0,0 +1,37 @@
...
+		"tags":  []any{"finance", "q3"},
+		"meta": map[string]any{
+			"author": "alice",
+		},
+		"count": 42, // non-string values are ignored
+	}
+
+	got := ExtractContent(value)
+	for _, want := range []string{"Quarterly Budget", "Review the numbers", "finance", "q3", "alice"} {
+		if !contains(got, want) {
```
sashankg: strings.Contains?

internal/forwarding/search_proxy.go
```diff
@@ -0,0 +1,20 @@
...
+	"net/http"
+	"net/http/httputil"
+	"net/url"
+)
+
+// NewSearchProxy returns a reverse proxy that forwards requests unchanged
+// (including the original Authorization header) to the search server at
+// searchHost. Unlike serviceProxy/pdsForwarding, no signing or auth
+// validation happens here — the search server validates the caller's token
+// itself via pear's existing org.getMetadata endpoint.
```
sashankg: no this is wrong. we should use regular service auth and the search server uses that to validate the user

sashankg (commented):
should be using sap to receive the records

internal/sap/resyncer.go
```diff
@@ -85,16 +89,16 @@ func (r *resyncer) dispatch(ctx context.Context) {
 	for {
 		rows, err := r.db.WithContext(ctx).Raw(`
 			UPDATE managed_repos SET state = 'resyncing'
-			WHERE rowid IN (
-				SELECT rowid FROM managed_repos
+			WHERE (space, did) IN (
```
arushibandi, line 92: rowid doesn't work for postgres, its a sqlite thing

internal/sap/resyncer.go
```diff
@@ -62,6 +62,10 @@ func (r *resyncer) run(ctx context.Context) {
 	for i := 0; i < r.parallelism; i++ {
 		go r.runWorker(ctx, i)
 	}
+	// Sweep for repos left pending/desynced/error by a prior process
+	// lifetime: nothing else will notify the dispatcher about them, since
+	// notifications only fire on new discovery or new live events.
+	r.dispatch(ctx)
```
arushibandi, line 68: if repos were in `pending` in startup, it needed another change to get picked up without this

internal/sap/outbox.go
```diff
@@ -76,6 +76,19 @@ func (m OutboxMessage) Ack(ctx context.Context) error {
 	return m.ack(ctx)
 }
 
+// NewOutboxMessageForTesting constructs an OutboxMessage for tests of code
+// outside this package that consumes Outbox — OutboxMessage's ack field is
+// unexported, so it can't otherwise be built with a working Ack from
+// another package.
+func NewOutboxMessageForTesting(
```
arushibandi: this is some testing thing, can make it better

sashankg: just move it to the test file, right?

sashankg (approved):

internal/sap/outbox.go
arushibandi: it's because of the unexported `ack` field

PR #426, opened 2026-06-23
[Authn] Validator

sashankg:
Callsites using the authn validator need to know if the credential is a user/org/instance cred so that they can use it for authz

### DIFF
--- a/internal/authn/auth_methods.go
+++ b/internal/authn/auth_methods.go
@@ -4,40 +4,76 @@ import (
 	"context"
 	"fmt"
 	"net/http"
+	"slices"
 
 	"github.com/bluesky-social/indigo/atproto/syntax"
 )
 
-type Method interface {
-	CanHandle(r *http.Request) bool
-	Validate(w http.ResponseWriter, r *http.Request, scopes ...string) (syntax.DID, bool)
-	ValidateRaw(ctx context.Context, token string, scopes ...string) (syntax.DID, bool, error)
+type CredentialType int
+
+const (
+	InstanceCredential CredentialType = iota
+	OrgCredential
+	UserCredential
+)
+
+type CredentialInfo struct {
+	Subject syntax.DID
+	Type    CredentialType
 }
 
-func Validate(
-	w http.ResponseWriter,
-	r *http.Request,
-	authMethod ...Method,
-) (syntax.DID, bool) {
-	for _, method := range authMethod {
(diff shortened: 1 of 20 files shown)


internal/authn/auth_methods.go
```diff
@@ -4,40 +4,76 @@ import (
...
+func NewValidator(authMethods ...Method) *Validator {
+	return &Validator{authMethods: authMethods}
+}
+
+func (v *Validator) WithSupportedCredentials(supportedCredentials ...CredentialType) *Validator {
+	v.supportedCredentials = supportedCredentials
+	return v
+}
+
+func (v *Validator) WithRequiredSubject() *Validator {
```
arushibandi: i don't get this

internal/authn/auth_methods.go
```diff
@@ -4,40 +4,76 @@ import (
...
+type Validator struct {
+	authMethods          []Method
+	supportedCredentials []CredentialType
+}
+
+func NewValidator(authMethods ...Method) *Validator {
+	return &Validator{authMethods: authMethods}
+}
+
+func (v *Validator) WithSupportedCredentials(supportedCredentials ...CredentialType) *Validator {
```
arushibandi: i think usually `WithXyz()` goes in the constructor on an `validatorOptions` type so you can't arbitrarily call this later in the Validator's lifetime for better safety

internal/server/server.go
```diff
@@ -64,7 +63,9 @@ func NewServer(
 
 // PutRecord puts a private record (see s.inner.putRecord)
 func (s *Server) PutRecord(w http.ResponseWriter, r *http.Request) {
-	callerDID, ok := authn.Validate(w, r, s.authMethods.oauth)
+	credInfo, ok := authn.NewValidator(s.authMethods.oauth).
```
arushibandi: do we have to create a new authn.Validator every time? can't we make one and reuse it?

arushibandi (approved):

internal/authn/auth_methods.go
sashankg: instance credentials don't have a specific user/org they are acting on behalf of. some endpoints require a specific user/org like `putRecord` so this is a convenience for those endpoints

internal/server/server.go
sashankg: no cuz we need to customize it per endpoint

PR #435, opened 2026-06-25
[DocsV2] Init

sashankg:
most changes are just api generations. typescript/apps/docs-server is main thing to look at. very rough and not scalable rn. will build it out as we go. at the very least this will let us write records that are searchable


lexicons/network/habitat/docs/crdt.json
```diff
@@ -0,0 +1,23 @@
+{
+    "lexicon": 1,
+    "id": "network.habitat.docs.crdt",
+    "defs": {
+        "main": {
+            "type": "record",
+            "description": "The CRDT (Yjs) state of a collaborative document. Each document is its own space; this record holds the canonical document state under the literal key 'self'.",
+            "key": "literal:self",
```
arushibandi, line 8: what does literal:self do? never seen this

lexicons/network/habitat/docs/createDoc.json
```diff
@@ -0,0 +1,45 @@
...
+                "schema": {
+                    "type": "object",
+                    "required": [
+                        "name"
+                    ],
+                    "properties": {
+                        "name": {
+                            "type": "string",
+                            "description": "The initial name of the document."
+                        }
```
arushibandi: i don't think this is required in our current doc / should be? unless it is for some technical reason?

lexicons/network/habitat/docs/markdown.json
```diff
@@ -0,0 +1,28 @@
+{
+    "lexicon": 1,
+    "id": "network.habitat.docs.markdown",
```
arushibandi, line 3: why not use `markpub.at`?

lexicons/network/habitat/docs/updateDoc.json
```diff
@@ -0,0 +1,50 @@
+{
+    "lexicon": 1,
+    "id": "network.habitat.docs.updateDoc",
```
arushibandi, line 3: can you document the flow somewhere? like to update a doc, does an editor also do a putRecord on their own space repo of their edits?

typescript/xrpc-openapi-gen/spec/api.json
```diff
@@ -848,6 +848,334 @@
         }
       }
     },
+    "/network.habitat.docs.createDoc": {
```
arushibandi, line 851: probably want to exclude `network.habitat.docs` in `api-docs` so those don't get added to the documentation

typescript/apps/docs-server/src/orgClient.ts
```diff
@@ -0,0 +1,182 @@
...
+  expiresAt: number;
+}
+
+// OrgClient holds the org's OAuth credential and makes authenticated XRPC calls
+// to pear on the org's behalf. It mirrors internal/sap/org_manager.go: a public
+// PKCE OAuth client (pear's OAuth server does not support confidential
+// client_assertion auth) that completes an authorization code flow once, then
+// auto-refreshes the access token. Every request carries
+// `Authorization: Bearer <token>` and `Habitat-Auth-Method: oauth`.
+export class OrgClient {
```
arushibandi, line 19: why are we doing this in ts btw? won't we want to use sap which will make things difficult because its in go?

typescript/apps/docs-server/src/render.ts
```diff
@@ -0,0 +1,175 @@
+import * as Y from "yjs";
+
+export interface RenderedDoc {
+  title: string;
+  markdown: string;
+}
+
+// renderDoc converts a TipTap/Yjs document (the "default" XML fragment) into
+// markdown and derives a title. It walks the Yjs types directly to avoid
+// pulling a ProseMirror/TipTap schema into the server.
```
arushibandi, line 10: why?

typescript/apps/docs-server/src/serviceAuth.ts
```diff
@@ -0,0 +1,106 @@
...
+
+interface ServiceAuthPayload {
+  iss: string;
+  aud: string;
+  exp: number;
+  lxm?: string;
+}
+
+// ServiceAuthVerifier fully verifies the service-auth JWT that pear signs on a
+// caller's behalf and forwards to us: it checks the audience (our DID), the
```
arushibandi, line 16: does service auth need to specify a `#service` in the did doc? like would we have to put `#habitat_docs` in there?

arushibandi (approved):

lexicons/network/habitat/docs/crdt.json
sashankg: https://atproto.com/specs/record-key#record-key-type-literal-value

lexicons/network/habitat/docs/markdown.json
sashankg: will need to write additional rendering code to translate it into that. will do in a follow up

lexicons/network/habitat/docs/updateDoc.json
sashankg: yeah we'll do that eventually. for now, it's just letting the app view publish the doc itself

typescript/apps/docs-server/src/orgClient.ts
sashankg: we need to handle yjs crdt diffs from the updateDocs endpoint.

typescript/apps/docs-server/src/serviceAuth.ts
sashankg: clarified offline

PR #437, opened 2026-06-25
[Relationships] Add lexicons

sashankg:


lexicons/network/habitat/relationship/defs.json
```diff
@@ -0,0 +1,107 @@
...
+                "did"
+            ],
+            "properties": {
+                "did": {
+                    "type": "string",
+                    "format": "did"
+                }
+            }
+        },
+        "groupSubject": {
```
arushibandi: do groups have admins / members? or is it just a simple user set

lexicons/network/habitat/relationship/group.json
```diff
@@ -0,0 +1,31 @@
+{
+    "lexicon": 1,
+    "id": "network.habitat.relationship.group",
+    "defs": {
+        "main": {
+            "type": "record",
+            "description": "A named group of subjects usable as a grantee in relationship tuples. Membership is expressed via network.habitat.relationship.tuple records (relation 'member', object = this group). Owned by the org repo within the space it governs.",
```
arushibandi: wdym "within the space it governs"?

is there one big space where all of these live? or are they defined on a per-space basis?

lexicons/network/habitat/group/profile.json
```diff
@@ -0,0 +1,31 @@
+{
+    "lexicon": 1,
+    "id": "network.habitat.group.profile",
+    "defs": {
+        "main": {
+            "type": "record",
+            "description": "Metadata for a group. A group is a space of type `network.habitat.group`; this profile record is the group-space's self record, holding its display name and description. Group membership is expressed as roles on the group-space (any role implies membership), and the group is used as a grantee elsewhere via a network.habitat.relationship.defs#spaceRoleSubject that references the group-space with role 'reader'.",
```
arushibandi: is there any way to define a space type? like description etc.?

lexicons/network/habitat/relationship/check.json
```diff
@@ -0,0 +1,55 @@
...
+    "defs": {
+        "main": {
+            "type": "query",
+            "description": "Check whether a user holds a role on a space, resolving through org/space-role usersets (groups are spaces, so group membership and nested groups resolve as space-role usersets) and built-in role implications (owner implies manager implies writer implies reader). Caller must have the reader role on the space.",
+            "parameters": {
+                "type": "params",
+                "required": [
+                    "did",
+                    "relation",
+                    "space"
```
arushibandi, line 13: i think this is right, but it feels like it obfuscates the group piece a bit (like then you have to fetch the group space then do this vs. looking up directly "is user in this group")

lexicons/network/habitat/relationship/defs.json
```diff
@@ -0,0 +1,79 @@
...
+            "properties": {
+                "did": {
+                    "type": "string",
+                    "format": "did"
+                }
+            }
+        },
+        "spaceRoleSubject": {
+            "type": "object",
+            "description": "All subjects holding a role on a space (a userset). Enables cross-space inheritance, e.g. spaceA's writers as writers of spaceB. Because groups are spaces, group membership is expressed as this userset over a group-space, e.g. role 'reader' meaning all members of the group.",
```
arushibandi: can you write this second sentence in more plain language if possible

lexicons/network/habitat/relationship/defs.json
arushibandi: i think this is answered by the change to make groups spaces

lexicons/network/habitat/relationship/listTuples.json
```diff
@@ -0,0 +1,85 @@
...
+                    "space": {
+                        "type": "string",
+                        "format": "uri",
+                        "description": "URI of the governing space whose tuples to list."
+                    },
+                    "object": {
+                        "type": "string",
+                        "format": "uri",
+                        "description": "Optional. Restrict to tuples whose object is this space or group URI."
+                    },
```
arushibandi, line 23: wonder if `subjectType` might be wanted here, where its either user did or group or another space role

arushibandi (commented):
i think i like it!

we need a readme /write up / examples / documentation for this for sure 

also a couple of my comments might be old because it changed while i was reviewing

arushibandi (approved):

lexicons/network/habitat/group/profile.json
sashankg: yeah they're going to be added to the lexicon spec: https://github.com/bluesky-social/proposals/blob/06d439e6be9004a086f392008e41acddd1a444ff/0016-permissioned-data/README.md#space-type-declarations

lexicons/network/habitat/relationship/check.json
sashankg: yeah we can have a separate "group" lexicon set. this is lower level and applies to any space type

PR #441, opened 2026-06-28
[Relationships] Implementation

sashankg:
Actual implementation of relationship xrpc endpoints. kept the implementations simple and dumb for now.

### DIFF
--- a/internal/fgastore/encoding.go
+++ b/internal/fgastore/encoding.go
@@ -25,6 +25,13 @@ func MemberUserString(did syntax.DID) string {
 	return "user:" + url.QueryEscape(did.String())
 }
 
+// SpaceUsersetString returns the FGA userset string for all subjects holding
+// `relation` on the given space, e.g. "space:<esc>#can_read". This is how a
+// space (including a group-space) is referenced as a grantee on another space.
+func SpaceUsersetString(uri habitat_syntax.SpaceURI, relation string) string {
+	return SpaceObjectKey(uri) + "#" + relation
+}
+
 // MemberUserToDID extracts a DID from an FGA user string.
 func MemberUserToDID(user string) (syntax.DID, error) {
 	if !strings.HasPrefix(user, "user:") {
--- a/internal/fgastore/encoding.go
+++ b/internal/fgastore/encoding.go
@@ -26,7 +26,7 @@ func MemberUserString(did syntax.DID) string {
 }
 
 // SpaceUsersetString returns the FGA userset string for all subjects holding
-// `relation` on the given space, e.g. "space:<esc>#can_read". This is how a
+// `relation` on the given space, e.g. "space:<spaceURI>#can_read". This is how a
 // space (including a group-space) is referenced as a grantee on another space.
 func SpaceUsersetString(uri habitat_syntax.SpaceURI, relation string) string {
 	return SpaceObjectKey(uri) + "#" + relation
--- a/internal/fgastore/fgastore_test.go
+++ b/internal/fgastore/fgastore_test.go
@@ -586,3 +586,42 @@ func TestEncodingHelpers_ReturnErrorsForInvalidInput(t *testing.T) {
 	_, err = ParseSpaceObjectKey("space:not-a-space-uri")
 	require.Error(t, err)
 }
+
+func TestCheck_SpaceUsersetCrossSpaceInheritance(t *testing.T) {
+	ctx := context.Background()
+	f := newTestSQLite(t)
+
+	// alice is a writer of space A.
+	require.NoError(t, f.Write(ctx, "user:alice", RelationSpaceWriter, "space:A"))
+	// all writers of A are writers of B (cross-space userset).
+	require.NoError(
+		t,
+		f.Write(ctx, "space:A#"+RelationSpaceWriter, RelationSpaceWriter, "space:B"),
+	)
+
+	ok, err := f.Check(ctx, "user:alice", RelationSpaceWriter, "space:B")
+	require.NoError(t, err)
+	require.True(t, ok, "writer of A should be writer of B")
+
+	// and by implication a reader of B.
+	ok, err = f.Check(ctx, "user:alice", RelationSpaceReader, "space:B")
+	require.NoError(t, err)
+	require.True(t, ok, "writer of B should imply reader of B")
+}
+
+func TestCheck_NestedGroupSpaces(t *testing.T) {
+	ctx := context.Background()
+	f := newTestSQLite(t)
+
+	// bob is a member (reader) of group-space A.
+	require.NoError(t, f.Write(ctx, "user:bob", RelationSpaceReader, "space:groupA"))
+	// group A's members are members of group B (nested groups).
+	require.NoError(
+		t,
+		f.Write(ctx, "space:groupA#"+RelationSpaceReader, RelationSpaceReader, "space:groupB"),
+	)
+
+	ok, err := f.Check(ctx, "user:bob", RelationSpaceReader, "space:groupB")
+	require.NoError(t, err)
+	require.True(t, ok, "member of nested group A should be a member of group B")
+}
--- a/internal/fgastore/fgastore_test.go
+++ b/internal/fgastore/fgastore_test.go
@@ -614,7 +614,7 @@ func TestCheck_NestedGroupSpaces(t *testing.T) {
 	f := newTestSQLite(t)
 
 	// bob is a member (reader) of group-space A.
-	require.NoError(t, f.Write(ctx, "user:bob", RelationSpaceReader, "space:groupA"))
+	require.NoError(t, f.Write(ctx, "user:bob", RelationSpaceOwner, "space:groupA"))
 	// group A's members are members of group B (nested groups).
 	require.NoError(
 		t,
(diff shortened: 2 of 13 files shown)


internal/fgastore/encoding.go
```diff
@@ -25,6 +25,13 @@ func MemberUserString(did syntax.DID) string {
 	return "user:" + url.QueryEscape(did.String())
 }
 
+// SpaceUsersetString returns the FGA userset string for all subjects holding
+// `relation` on the given space, e.g. "space:<esc>#can_read". This is how a
```
arushibandi: what is `<esc>`

internal/fgastore/fgastore_test.go
```diff
@@ -586,3 +586,42 @@ func TestEncodingHelpers_ReturnErrorsForInvalidInput(t *testing.T) {
 	_, err = ParseSpaceObjectKey("space:not-a-space-uri")
 	require.Error(t, err)
 }
+
+func TestCheck_SpaceUsersetCrossSpaceInheritance(t *testing.T) {
```
arushibandi, line 590: can we test some other combinations too? i think our tests should be very comprehensive here. thinking:
- test owner / manage_permissions relations with nesting / cross space
- test deleting a relation and correctness on nesting / cross space implications

internal/fgastore/model.go
```diff
@@ -112,24 +135,16 @@ func authModel() *openfgav1.AuthorizationModel {
 				Metadata: &openfgav1.Metadata{
 					Relations: map[string]*openfgav1.RelationMetadata{
 						RelationSpaceOwner: {
-							DirectlyRelatedUserTypes: []*openfgav1.RelationReference{
-								{Type: TypeUser},
-							},
+							DirectlyRelatedUserTypes: spaceDirectlyRelatedUserTypes(),
```
arushibandi, line 138: can you explain. i don't understand what this is doing (what does `DirectlyRelatedUserTypes` do in openFGA)

internal/relationship/relationship.go
```diff
@@ -0,0 +1,197 @@
+// Package relationship implements app-managed ReBAC over spaces. It mirrors
+// relationship tuples (subject, relation, object) into both AT Protocol records
+// in the governing space (so they are org-owned and readable by other apps) and
+// the OpenFGA graph (for Check/expand).
+//
+// Groups, and an org's member/admin sets, are modeled as spaces: there is no
+// dedicated group or organization concept here. A group is a space (type
+// network.habitat.group) and is referenced as a grantee via a SpaceRoleSubject
```
arushibandi, line 8: is there any way to enforce this ? can any app set groups?

internal/relationship/relationship.go
```diff
@@ -0,0 +1,197 @@
...
+// inheritance.
+type Subject interface {
+	isSubject()
+	// Kind reports the concrete subject kind.
+	Kind() SubjectKind
+	// fgaUserString returns the OpenFGA user string for the subject.
+	fgaUserString() (string, error)
+	// toInterface serializes the subject to the lexicon union map form stored in
+	// the AT Protocol record and returned over XRPC.
+	toInterface() map[string]any
```
arushibandi, line 87: nit: `toLexicon`?

internal/relationship/relationship.go
```diff
@@ -0,0 +1,197 @@
...
+	return map[string]any{
+		"$type": subjectTypeSpaceRole,
+		"space": s.Space.String(),
+		"role":  string(s.Role),
+	}
+}
+
+// ParseSubject decodes a subject from the generated union interface{} value
+// (as produced by JSON decoding of an XRPC body or CBOR decoding of a record).
+func ParseSubject(generic any) (Subject, error) {
```
arushibandi, line 136: can this and `parseObject` take in `map[string]any`? or does the caller param really have `any` type

internal/relationship/server.go
```diff
@@ -0,0 +1,408 @@
...
+	}
+	if !authorized {
+		http.Error(w, "not authorized", http.StatusForbidden)
+		return false
+	}
+	return true
+}
+
+func (s *Server) WriteTuple(w http.ResponseWriter, r *http.Request) {
+	credInfo, ok := authn.NewValidator(
```
arushibandi, line 86: why do we need `NewValidator` if every callsite of this is `NewValidator(..).Validate(..)`. can we not condense this into one function `Validate`?

internal/relationship/server.go
```diff
@@ -0,0 +1,408 @@
...
+		return
+	} else if err != nil {
+		utils.LogAndHTTPError(r.Context(), w, err, "write tuple", http.StatusInternalServerError)
+		return
+	}
+
+	s.writeJSON(w, r, habitat.NetworkHabitatRelationshipWriteTupleOutput{Uri: uri.String()})
+}
+
+func (s *Server) DeleteTuple(w http.ResponseWriter, r *http.Request) {
```
arushibandi, line 131: one-line comments on all the functions would be apprecaited

internal/relationship/server.go
arushibandi: or store a `NewValidator` on the server? idk this pattern feels wrong

internal/relationship/server.go
```diff
@@ -0,0 +1,408 @@
...
+	}
+
+	// Only return spaces the caller is allowed to read.
+	out := make([]string, 0, len(spaceURIs))
+	for _, space := range spaceURIs {
+		readable, err := s.authorize(
+			r.Context(),
+			credInfo.Subject,
+			space,
+			fgastore.RelationSpaceReader,
```
arushibandi, line 341: does `store.ListObjects` not enforce this ipmplicitly?

internal/relationship/store_test.go
```diff
@@ -0,0 +1,348 @@
...
+
+// flakyFGA wraps a real FGA store but can be told to fail mutating WriteRaw
+// calls, to exercise transaction rollback.
+type flakyFGA struct {
+	fgastore.Store
+	failWrites bool
+}
+
+func (f *flakyFGA) WriteRaw(ctx context.Context, req *openfgav1.WriteRequest) error {
+	if f.failWrites {
```
arushibandi, line 299: can you make this happen according to a % to make it randomly flaky ? and then run all the tests on it?

internal/relationship/store.go
```diff
@@ -0,0 +1,361 @@
...
+		return "", err
+	}
+	fgaUser, err := subject.fgaUserString()
+	if err != nil {
+		return "", err
+	}
+
+	// Idempotency: if an identical tuple record already exists, reuse it so we
+	// don't accumulate duplicate records that map to a single FGA tuple (which
+	// would desync on delete).
```
arushibandi, line 84: can we do this via the database instead of manually checking?

internal/relationship/store.go
```diff
@@ -0,0 +1,361 @@
...
+						tuple.NewTupleKey(fgastore.SpaceObjectKey(object), fgaRelation, fgaUser),
+					),
+				},
+				OnMissing: "ignore",
+			},
+		})
+	})
+}
+
+// ListTuples returns the tuples governing a space, applying optional filters.
```
arushibandi, line 184: could probably filter this in the db query right?

arushibandi (approved):

internal/fgastore/model.go
sashankg: it's just how you define what's allowed to be included as subjects for these relationships

internal/relationship/relationship.go
sashankg: no only the admin app will be granted scopes to modify groups

internal/relationship/relationship.go
sashankg: i agree toInterface is a bad name but `toLexicon` is also confusing. it's not returning a lexicon

internal/relationship/relationship.go
sashankg: the lexicon defines it as a union so go just treats it as any

internal/relationship/server.go
sashankg: We're allowing each endpoint to specify its own supported auth methods and credential types. For example, in spaces, you can't use an instance credential to putRecord but you could use it to getRecord. We need to build a validator that meets the requirements of the endpoint. I think this is a common pattern

internal/relationship/server.go
sashankg: seems redundant on the service methods since their interace is defined in the lexicon description and the implementation is defined in the underlying store. The server methods just parse inputs and encode outputs

internal/relationship/server.go
sashankg: no. the did we pass to ListObjects is a request param so the caller can pass in anybody's did. We need to enforce that the returned spaces are granted to the caller.

internal/relationship/store_test.go
sashankg: that's more of an integration test than a unit test. will start adding to the integration tests

internal/relationship/store.go
sashankg: yeah maybe but just doing the dumbest fool proof thing for now. will improve later

internal/relationship/store.go
sashankg: same as above just doing the simplest thing for now. it is slightly harder to do this in the query since we'd have to parse the record

PR #442, opened 2026-06-28
[Password] Fix new member password

sashankg:


arushibandi (approved):

arushibandi:
i also think we should support an email + confirmation code method for login (i suspect this is what some orgs want), maybe that can replace password. and in the future wonder if we need to support any login method i.e the org can let members decide.

PR #443, opened 2026-06-28
[OAuthClient] Refactor out of sap

sashankg:


internal/oauthclient/gormstore.go
```diff
@@ -0,0 +1,136 @@
...
+	"fmt"
+	"time"
+
+	"github.com/bluesky-social/indigo/atproto/auth/oauth"
+	"github.com/bluesky-social/indigo/atproto/syntax"
+	"gorm.io/gorm"
+)
+
+// sessionRow stores oauth.ClientSessionData keyed by (did, session_id).
+type sessionRow struct {
```
arushibandi, line 15: do you need `gorm.Model` on this?

arushibandi (approved):

internal/oauthclient/gormstore.go
sashankg: gorm.Model just adds an ID and created, updated timestamps. we don't want an ID since State is the primary column

PR #446, opened 2026-06-30
[Validator] Include org in credInfo

sashankg:

### DIFF
--- a/internal/core/org.go
+++ b/internal/core/org.go
@@ -1,0 +1,44 @@
+package core
+
+import (
+	"context"
+	"time"
+
+	"github.com/bluesky-social/indigo/atproto/identity"
+	"github.com/bluesky-social/indigo/atproto/syntax"
+	"github.com/habitat-network/habitat/api/habitat"
+	"github.com/habitat-network/habitat/internal/db"
+)
+
+// Org represents a single organization on a pear instance.
+type Org interface {
+	DID() syntax.DID
+
+	AddAdmin(ctx context.Context, admin syntax.DID) error
+	GetAdmins(ctx context.Context) ([]syntax.DID, error)
+	GetMembers(ctx context.Context) ([]syntax.DID, error)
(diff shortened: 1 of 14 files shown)


internal/org/login_router.go
```diff
@@ -16,8 +17,15 @@ type LoginRouter struct {
...
-func (r *LoginRouter) getProvider(org Org) login.Provider {
-	switch org.loginMethod(context.Background() /* todo: fix context */) {
+func (r *LoginRouter) getProvider(org core.Org) login.Provider {
+	type loginMethodProvider interface {
+		loginMethod(ctx context.Context) loginMethod
+	}
+	p, ok := org.(loginMethodProvider)
+	if !ok {
+		return nil
+	}
```
arushibandi: :/

why can't you make this a method on org directly ?

arushibandi: you should just make this a public interface that org implements ? i don't think we should be type asserting

internal/core/org.go
```diff
@@ -0,0 +1,44 @@
...
+	"context"
+	"time"
+
+	"github.com/bluesky-social/indigo/atproto/identity"
+	"github.com/bluesky-social/indigo/atproto/syntax"
+	"github.com/habitat-network/habitat/api/habitat"
+	"github.com/habitat-network/habitat/internal/db"
+)
+
+// Org represents a single organization on a pear instance.
```
arushibandi, line 21: i don't really get the need for this right now but sure

internal/org/login_router.go
sashankg: oh my bad AI must have done that. yeah will fix

internal/core/org.go
sashankg: if i don't move it to core it creates a dep cycle cuz org/server needs authn and authn needs the Org interface.

arushibandi (approved):

PR #450, opened 2026-06-30
[Home] Initial setup

sashankg:
add lexicon and initial implementation of home server


lexicons/network/habitat/groups/updateGroup.json
```diff
@@ -0,0 +1,59 @@
+{
+    "lexicon": 1,
+    "id": "network.habitat.groups.updateGroup",
+    "defs": {
+        "main": {
+            "type": "procedure",
+            "description": "Update a group's profile (name and/or description). Caller must be able to manage the group.",
+            "input": {
+                "encoding": "application/json",
+                "schema": {
```
arushibandi, line 10: nit: `updateGroupMetadata`?

lexicons/network/habitat/groups/addMember.json
```diff
@@ -0,0 +1,62 @@
+{
+    "lexicon": 1,
+    "id": "network.habitat.groups.addMember",
```
arushibandi, line 3: editMembers? or add a deleteMember?

cmd/home/flags.go
```diff
@@ -0,0 +1,53 @@
...
+		&cli.StringFlag{
+			Name:    fSecret,
+			Usage:   "Secret used in the OAuth client-metadata signing",
+			Value:   "secret",
+			Sources: cli.EnvVars("HOME_SECRET"),
+		},
+		&cli.StringFlag{
+			Name:    fOrgHandle,
+			Usage:   "Handle of the org this home server manages groups for",
+			Value:   "acmecorp.pear.local.habitat.network",
```
arushibandi, line 43: i don't think it needs this ? should be multi-tenant, this is something we are hosting

cmd/home/groups.go
```diff
@@ -0,0 +1,332 @@
...
+
+// GroupService implements the network.habitat.groups.* endpoints. It reads
+// group membership from the sap-fed index (Store) and performs writes against
+// pear using the org credential held by the oauth client.
+type GroupService struct {
+	store    *Store
+	oauthApp *oauthclient.App
+}
+
+func NewGroupService(store *Store, oauthApp *oauthclient.App) *GroupService {
```
arushibandi, line 35: why build this as a separate service instead of directly into pear?

cmd/home/pearclient.go
```diff
@@ -0,0 +1,174 @@
...
+	habitat_syntax "github.com/habitat-network/habitat/internal/syntax"
+)
+
+// GroupSpaceType is the space type every group is created under. The
+// network.habitat.group.profile self record holds the group's metadata.
+const GroupSpaceType = "network.habitat.group"
+
+const (
+	collectionGroupProfile = "network.habitat.group.profile"
+	collectionTuple        = "network.habitat.relationship.tuple"
```
arushibandi, line 23: should these go in pear as well? can anyone write these or just the groups service? bc we don't want people writing invalid records with these collections right?

arushibandi (approved):

cmd/home/groups.go
sashankg: yeah sap's syncing rn is tied to the network boundary. we can abstract that out so that it can sync within pear as well and then we can merge this in. this is just the easiest way with what we have so far

cmd/home/flags.go
sashankg: yeah will fix in a follow up. was just simplest way to get started

cmd/home/pearclient.go
sashankg: it will be scoped so that only authorized apps can and we can only allow this app to be authorized

PR #451, opened 2026-07-01
[OrgStore] Move identity creation to store

sashankg:
- **[OrgStore] Move identity creation to store**
- **simplify struct**

### DIFF
--- a/internal/org/store.go
+++ b/internal/org/store.go
@@ -104,6 +104,7 @@ func (s *storeImpl) orgFromModel(org *organization) (*orgImpl, error) {
 		db:              s.db,
 		handleSubdomain: org.HandleSubdomain,
 		name:            org.Name,
+		method:          org.LoginMethod,
 	}, nil
 }
 
--- a/internal/org/store.go
+++ b/internal/org/store.go
@@ -10,11 +10,19 @@ import (
 
 	"github.com/bluesky-social/indigo/atproto/identity"
 	"github.com/bluesky-social/indigo/atproto/syntax"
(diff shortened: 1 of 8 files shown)


internal/org/store.go
```diff
@@ -251,3 +268,160 @@ func (s *storeImpl) GetMemberByLoginID(ctx context.Context, loginID string) (*Me
 		LoginID: m.LoginID,
 	}, nil
 }
+
+func (s *storeImpl) ValidateAdminSignedToken(
```
arushibandi, line 273: seems like comments got lost when these got copied over :/

internal/org/store.go
```diff
@@ -74,19 +99,11 @@ func NewStore(
...
-		hive:             s.hive,
-		db:               s.db,
-		signingSecret:    signingSecret,
-		handleSubdomain:  org.HandleSubdomain,
-		method:           org.LoginMethod,
-		passwordProvider: s.passwordProvider,
-		name:             org.Name,
+		orgID:           org.ID,
+		db:              s.db,
+		handleSubdomain: org.HandleSubdomain,
```
arushibandi, line 105: do we even need this? can the caller just use `org` directly?

arushibandi (approved):
nice change some nits

internal/org/store.go
sashankg: yeah rn we have `orgImpl` which implements core.Org and `organization` which is the gorm model. we can probably merge them but will leave to a follow up

PR #452, opened 2026-07-01
Add /proxy XRPC endpoint to sap internal server

sashankg:
## Motivation

This lets users of `sap` interact with `pear` **without managing their own auth sessions**. `sap` already tracks OAuth sessions for the orgs it manages (crawling their repos, subscribing to their firehose). This endpoint reuses those sessions so a caller can proxy an XRPC request through `sap` and have it authenticated on the org's behalf — instead of every caller running its own OAuth flow and refreshing tokens.

## What changed

**`internal/sap/sap.go`** — added `Sap.GetClient(ctx, did)`. It resolves the managed org's tracked OAuth session (DID → `sessionID`, which only `sap` knows) and returns an authenticated `*http.Client` via the existing `oauthclient.App`. The client's transport resolves relative paths against the org's Habitat host and attaches the access token — the same path the crawler/resyncer already use.

**`cmd/sap/server.go`** — added `handleProxy`:
- Reads the target DID from the `Habitat-Did` request header (`400` if missing/invalid).
- Resolves it to the tracked session via `Sap.GetClient` (`502` if there's no session for that DID).
- Rewrites the path: everything after `/proxy/` becomes `/xrpc/<method>`, preserving the original HTTP method, query string, headers, and body.
- Clones the caller's headers, scrubs hop-by-hop headers and the `Habitat-Did`/`Authorization` headers we replace, then forwards. The OAuth transport attaches the bearer token.
- Streams pear's status, headers, and body straight back.

**`cmd/sap/main.go`** — registered `/proxy/` on the **internal** mux (alongside `/health`, `/org/*`, `/channel`), keeping it off the public OAuth port so access can be restricted to trusted callers.

**`cmd/sap/proxy_test.go`** — tests covering the happy path (method/query/body/token forwarded, `/xrpc/` rewrite, `Habitat-Did` stripped, response passed through) plus missing-header (`400`) and unknown-DID (`502`) cases.

## Usage

```
POST /proxy/network.habitat.space.createRecord?collection=note
Habitat-Did: did:plc:someorg
<body>
```

is forwarded to that org's pear as `POST /xrpc/network.habitat.space.createRecord?collection=note` with the org's OAuth token attached.

## Testing

- `go test ./...` in the `cmd/sap` module — pass
- `go build` / `go vet` on `cmd/sap` and `internal/sap` — clean
- `golangci-lint run` on both — 0 issues

🤖 Generated with [Claude Code](https://claude.com/claude-code)

---
_Generated by [Claude Code](https://claude.ai/code/session_01DVvUZodx7VwMmxweE5eLFP)_

### DIFF
--- a/cmd/sap/server.go
+++ b/cmd/sap/server.go
@@ -3,13 +3,26 @@ package main
 import (
 	"encoding/json"
 	"fmt"
+	"io"
 	"log/slog"
 	"net/http"
+	"net/url"
+	"strings"
 
 	"github.com/bluesky-social/indigo/atproto/syntax"
 	"github.com/habitat-network/habitat/internal/oauthclient"
 	"github.com/habitat-network/habitat/internal/sap"
 )
+
+// habitatDIDHeader names the DID the caller wants the proxied request to be
+// authenticated as. sap looks up the OAuth session it tracks for this DID.
+const habitatDIDHeader = "Habitat-Did"
+
+// hopByHopHeaders are connection-scoped and must not be forwarded to pear per
+// the HTTP/1.1 spec (RFC 7230 §6.1).
+var hopByHopHeaders = []string{
+	"Connection", "Transfer-Encoding", "Te", "Upgrade", "Keep-Alive",
+}
 
 type server struct {
 	sap         *sap.Sap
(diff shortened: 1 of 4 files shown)


cmd/sap/server.go
```diff
@@ -94,6 +107,82 @@ func (s *server) handleOAuthCallback(w http.ResponseWriter, r *http.Request) {
...
+// Habitat-Did header. The path after /proxy/ is the XRPC method, forwarded to
+// pear as /xrpc/<method> with the original method, query params, headers, and
+// body preserved.
+func (s *server) handleProxy(w http.ResponseWriter, r *http.Request) {
+	didStr := r.Header.Get(habitatDIDHeader)
+	if didStr == "" {
+		http.Error(w, "missing "+habitatDIDHeader+" header", http.StatusBadRequest)
+		return
+	}
+	did, err := syntax.ParseDID(didStr)
```
arushibandi, line 121: how is the caller authenticated though? or I guess this is only exposed internally in the app server?

arushibandi (approved):

cmd/sap/server.go
sashankg: yeah it's on the internal port so it's the callers responsibility to keep it private or add some auth in front of it

PR #456, opened 2026-07-02
[Org] Add memberships to fga store

sashankg:
Encode the org membership/admin info into the fga store so that we can create an org space that can be used to grant access to the entire org

-------------------------
- master
  - **[Org] Add memberships to fga store** :point_left:
    - #458


internal/org/org.go
```diff
@@ -122,6 +133,14 @@ func (s *orgImpl) DowngradeAdmin(ctx context.Context, admin syntax.DID) error {
 	if adminCount < 2 {
 		return ErrLastAdmin
 	}
+	if err := s.fga.Delete(
+		ctx,
+		fgastore.MemberUserString(admin),
+		fgastore.RelationAdmin,
+		fgastore.OrgObjectKey(s.orgID),
+	); err != nil {
+		return err
```
arushibandi, line 180: do you need to add RelationMember ?

arushibandi: otherwise if you have someone who 1) creates the org 2) makes someone else an admin 3) that other person downgrades them, i think they will no longer have member relation

you might also wnat to add a test for that sequence ^ / generate other similar sequences

arushibandi (approved):

internal/org/org.go
sashankg: yeah good catch

PR #457, opened 2026-07-02
[Relationship] Support space roles in check

sashankg:
Update the check xrpc endpoint to support checking if a space role has permission to another space. This would let callers be able to check if the entire org has permission by doing `subject=ats://<org>/network.habitat.organization/self&subjectRole=reader`


internal/relationship/store_test.go
```diff
@@ -238,15 +274,58 @@ func TestGroupAsSpace(t *testing.T) {
...
-	allowed, err = rel.Check(t.Context(), bob, RoleWriter, target)
+	allowed, err = rel.Check(t.Context(), UserSubject{DID: bob}, RoleWriter, target)
 	require.NoError(t, err)
 	require.False(t, allowed, "non-member should not have access")
 }
 
+// TestCheck_SpaceSubject checks a space-role userset directly as the subject of
+// a Check (rather than an individual user), resolving role implications and
+// distinguishing usersets that were never granted.
+func TestCheck_SpaceSubject(t *testing.T) {
```
arushibandi, line 340: can you also check the organization case that you mention in the PR description?

arushibandi (approved):

PR #458, opened 2026-07-02
[Relationships] Add org space

sashankg:
add an `ats://<org-did>/network.habitat.organization/self` space that clients can use to grant grant their spaces to the entire org. it's an implicit space for now since there are no records in it. might change later


internal/relationship/server.go
```diff
@@ -52,6 +53,7 @@ func (s *Server) authorize(
 		relation,
 		fgastore.SpaceObjectKey(space),
 		ownerContextualTuple(space),
+		fgastore.OrgMemberContextualTuple(callerOrg),
```
arushibandi: can't you just construct the callerOrg here instead of having to take it in? then you don't have to change all those callsites? makes it clearer imo

arushibandi (approved):

internal/relationship/server.go
sashankg: construct from what? like call get org again? shouldn't have to call get org multiple times in single request

arushibandi: like change it to `OrgMemberContextualTuple(callerDID)`

that way all the `authorize` calls don't have to change

arushibandi: it seems like everywhere you supply callerOrg it is just did.Org.DID(), so might as well just do that once here

sashankg: it's `credInfo.Org.DID()` . I guess I can pass the whole credInfo down. will fix

PR #459, opened 2026-07-02
[Frontend] Add collections page

sashankg:
Add collections page. On listCollections, we fetch the spaces the user has permission to, then construct the response based on that. this makes sure we're only returning ones the user has permission to see


lexicons/network/habitat/collections/listRecords.json
```diff
@@ -0,0 +1,41 @@
+{
+    "lexicon": 1,
+    "id": "network.habitat.collections.listRecords",
```
arushibandi, line 3: can you delete `network.habitat.repo.listRecords`

cmd/home/store.go
```diff
@@ -66,7 +92,7 @@ type Store struct {
 }
 
 func NewStore(db *gorm.DB) (*Store, error) {
-	if err := db.AutoMigrate(&groupRow{}, &tupleRow{}, &orgSessionRow{}); err != nil {
+	if err := db.AutoMigrate(&groupRow{}, &tupleRow{}, &recordRow{}, &orgSessionRow{}); err != nil {
```
arushibandi, line 95: no automigrate for collectionCount?

arushibandi: oh wait you're just fetching this not indexing it

cmd/home/collections_test.go
```diff
@@ -0,0 +1,151 @@
...
+		syntax.NSID(collection),
+		syntax.RecordKey(rkey),
+	)
+}
+
+const (
+	space1 = "ats://did:web:org/network.habitat.space/spaceone"
+	space2 = "ats://did:web:org/network.habitat.space/spacetwo"
+)
+
```
arushibandi, line 27: maybe test that if you have space inheritance (readers of one space are readers of another), the counts are still correct

arushibandi (approved):

lexicons/network/habitat/collections/listRecords.json
sashankg: we still use it in the docs app

cmd/home/collections_test.go
sashankg: that should be verified by the listSpaces implementation. we don't do any permission resolution in home. just ask what the user has permission to see and return those

PR #460, opened 2026-07-03
[Org]: Add contact email to org creation flow

arushibandi:
Orgs now provide a required contact email on creation so Habitat can reach out to the person setting up the org. Threaded through the network.habitat.org.create lexicon, the organization store/model, and the create-org form (pre-filled from the Google login email when that login method is used).

### DIFF
--- a/frontend/src/routes/org/create.tsx
+++ b/frontend/src/routes/org/create.tsx
@@ -1,6 +1,7 @@
 import {
   Button,
   Field,
+  FieldDescription,
   FieldError,
   FieldLabel,
   Input,
@@ -27,6 +28,7 @@ interface FormValues {
   admin_password: string;
   login_method: "password" | "atproto" | "google";
   login_id: string;
+  contact_email: string;
   handle_subdomain: string;
   use_custom_instance: boolean;
   custom_domain: string;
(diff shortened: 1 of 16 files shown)


frontend/src/routes/org/create.tsx
```diff
@@ -99,6 +102,15 @@ function CreateOrgPage() {
   const loginMethod = watch("login_method");
   const useCustomInstance = watch("use_custom_instance");
   const customDomain = watch("custom_domain");
+  const googleLoginId = watch("login_id");
+
+  const [contactEmailTouched, setContactEmailTouched] = useState(false);
+
+  useEffect(() => {
```
sashankg, line 109: lol whatever's going on in this file is insane. i'll clean it up after you merge this

sashankg (approved):

internal/org/models.go
```diff
@@ -15,6 +15,7 @@ type organization struct {
 	SigningSecret   string           // base64-encoded HMAC-SHA256 key for invite tokens
 	CreatedAt       time.Time
 	HandleSubdomain string `gorm:"unique"`
+	ContactEmail    string `gorm:"not null"` // email for reaching out to the org about its account
```
sashankg: `not null` will probably cause an issue for existing orgs. let's maybe add `unique` though

arushibandi: oh true

PR #466, opened 2026-07-06
[DocsV2] Use sap and relationship endpoints

sashankg:


arushibandi (approved):
didn't do the closest read but for demo purposes seems good

PR #610, opened 2026-07-23
[DB] Single automigrate call

sashankg:


sashankg:
looks like automigrate still makes multiple db calls so this doesn't help startup time

PR #729, opened 2026-08-08
[cleanup]: move site stuff to other repo; delete unused

arushibandi:
Deleting / .gitignoring a bunch of folders that are unused or unnecessary for the public-facing repo. Moving all website related stuff (site code + waitlist service) to a new private repo. Will test that out and then merge.


sashankg (commented):
I'm a big believer in keeping everything in one monorepo including the website. For example, if we want to reuse the design system UI components between the web app and the website we'd have to factor that out and manage versioning. What's the reason for moving it to a private repo? all the website content is public anyway

arushibandi:
> I'm a big believer in keeping everything in one monorepo including the website. For example, if we want to reuse the design system UI components between the web app and the website we'd have to factor that out and manage versioning. What's the reason for moving it to a private repo? all the website content is public anyway

Since we're open-sourcing the habitat core platform, we need to put all of that in one repo that only does that. If that's this repo, we should move habitat-the-company specific stuff out.

sashankg:
Ok then whatever private repo we're moving stuff to should also be a monorepo

sashankg (approved):

arushibandi:
> Ok then whatever private repo we're moving stuff to should also be a monorepo

yes i agree. i can rename it as such

PR #737, opened 2026-08-11
[Spaces] Fix space credential kid

sashankg:


internal/authn/testutil/stub.go
```diff
@@ -97,3 +97,23 @@ func (s *failure) Validate(
...
+	w http.ResponseWriter,
+	r *http.Request,
+	scopes ...string,
+) (*authn.CredentialInfo, bool) {
+	return nil, false
+}
+
+func NewCantHandleMethod() authn.Method {
+	return &cantHandle{}
+}
```
arushibandi, line 119: is this even used? also where is the test for this bug?

PR #830, opened 2026-08-18
[perms]: use new perms store

arushibandi:
Use this in all the easy swap ins; everything except for org server and relationship lexicon server

### DIFF
--- a/internal/perms/store.go
+++ b/internal/perms/store.go
@@ -64,6 +64,7 @@ type Store interface {
 		did syntax.DID,
 		space habitat_syntax.SpaceURI,
 		role habitat_syntax.SpaceRole,
+		belongsToOrg syntax.DID,
 	) (bool, error)
 	CheckSpaceRelationHasSpaceRole(
 		ctx context.Context,
@@ -335,13 +336,14 @@ func (s *store) UnsafeRevokeAllSpaceRoles(
 }
 
 // CheckUserHashabitat_syntax.SpaceRole implements [Store]. The space's owner is treated as
-// an implicit habitat_syntax.SpaceRoleOwner, and the owning org's members as implicit
-// habitat_syntax.SpaceRoleReaders, without either needing a stored tuple.
+// an implicit habitat_syntax.SpaceRoleOwner, and members of belongsToOrg (the caller's own
+// org) as implicit habitat_syntax.SpaceRoleReaders, without either needing a stored tuple.
 func (s *store) CheckUserHasSpaceRole(
 	ctx context.Context,
 	did syntax.DID,
 	space habitat_syntax.SpaceURI,
 	role habitat_syntax.SpaceRole,
+	belongsToOrg syntax.DID,
 ) (bool, error) {
 	return s.fga.Check(
 		ctx,
@@ -349,7 +351,7 @@ func (s *store) CheckUserHasSpaceRole(
 		fgaRelationFromRole[role],
 		fgastore.SpaceObjectKey(space),
 		fgastore.OwnerContextualTuple(space),
-		fgastore.OrgMemberContextualTuple(space.SpaceOwner()),
+		fgastore.OrgMemberContextualTuple(belongsToOrg),
 	)
 }
 
--- a/internal/perms/store.go
+++ b/internal/perms/store.go
@@ -111,7 +111,8 @@ func (s *store) AddUserRelation(
 			/* object is the space being written into itself */
 		}
 		var err error
-		uri, _, err = s.spaces.WithTx(tx).PutRecord(ctx /* TODO should i use tx.Statement.Context */, space, space.SpaceOwner(), habitat_syntax.UserRelationCollection, "" /* rkey should be generated */, record)
+		uri, _, err = s.spaces.WithTx(tx).
+			PutRecord(ctx /* TODO should i use tx.Statement.Context */, space, space.SpaceOwner(), habitat_syntax.UserRelationCollection, "" /* rkey should be generated */, record)
 		if err != nil {
 			return fmt.Errorf("err putting relationship record: %w", err)
 		}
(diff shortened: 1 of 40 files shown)


internal/perms/store.go
```diff
@@ -271,51 +295,34 @@ func (s *store) CheckSpaceRelationHasSpaceRole(
...
 	space habitat_syntax.SpaceURI,
-) ([]syntax.DID, []SpaceRoleSubject, error) {
-	tuples, err := s.fga.Read(ctx, fgastore.Tuple{Object: fgastore.SpaceObjectKey(space)})
+) ([]syntax.DID, error) {
+	users, err := s.fga.ListUsers(
+		ctx,
+		fgastore.SpaceObjectKey(space),
+		fgastore.RelationSpaceReader,
+		fgastore.OwnerContextualTuple(space),
+		fgastore.OrgMemberContextualTuple(space.SpaceOwner()),
```
sashankg: this is not right. the org id passed in here needs to be the org the user belongs to. this would make fga store think the user is always in the same org as the space owner we're checking against

internal/syntax/relationship.go
```diff
@@ -0,0 +1,18 @@
...
+// SpaceRole is an access-control role held on a space. The hierarchy
+// (owner ⇒ manager ⇒ writer ⇒ reader) is enforced by the OpenFGA model
+// (internal/fgastore's authModel), not by this package.
+type SpaceRole string
+
+const (
+	SpaceRoleOwner   SpaceRole = "owner"
+	SpaceRoleManager SpaceRole = "manager"
+	SpaceRoleWriter  SpaceRole = "writer"
+	SpaceRoleReader  SpaceRole = "reader"
```
arushibandi, line 14: these are separate from what's in the fga model fyi

internal/perms/store.go
```diff
@@ -80,35 +63,33 @@ type Store interface {
...
 	ListObjects(ctx context.Context, did syntax.DID) ([]habitat_syntax.SpaceURI, error)
 }
 
+type TxRunner interface {
+	Transaction(fc func(tx *gorm.DB) error, opts ...*sql.TxOptions) (err error)
+}
+
 type store struct {
-	fga fgastore.Store
+	db     TxRunner
```
sashankg: why not just gorm.DB?

internal/perms/store.go
```diff
@@ -118,92 +99,191 @@ func (s *store) AddUserRelation(
...
+	err := s.db.Transaction(func(tx *gorm.DB) error {
+		record := map[string]any{
+			"subject":   did.String(),
+			"relation":  string(role),
+			"createdAt": time.Now().UTC().Format(time.RFC3339),
+			/* object is the space being written into itself */
+		}
+		var err error
+		uri, _, err = s.spaces.WithTx(tx).
+			PutRecord(ctx /* TODO should i use tx.Statement.Context */, space, space.SpaceOwner(), habitat_syntax.UserRelationCollection, "" /* rkey should be generated */, record)
```
sashankg: I think ctx is fine

internal/perms/store.go
```diff
@@ -118,92 +99,191 @@ func (s *store) AddUserRelation(
...
-					fgastore.SpaceObjectKey(object),
-					fgaRelationFromRole[objectRole],
-					fgastore.SpaceUsersetString(subject, fgaRelationFromRole[subjectRole]),
-				),
+	objectRole habitat_syntax.SpaceRole,
+) (habitat_syntax.SpaceRecordURI, error) {
+
+	var uri habitat_syntax.SpaceRecordURI
+	err := s.db.Transaction(func(tx *gorm.DB) error {
+		record := map[string]any{
```
sashankg, line 179: should we use the generated type here? I know we've only used the generated types in servers so far but would be nice to keep this type safe

internal/perms/store.go
```diff
@@ -118,92 +99,191 @@ func (s *store) AddUserRelation(
...
+			"subject":     subject.String(),
+			"subjectRole": string(subjectRole),
+			"relation":    string(objectRole),
+			"createdAt":   time.Now().UTC().Format(time.RFC3339),
+			/* object is the space being written into itself */
+		}
+		var err error
+		uri, _, err = s.spaces.WithTx(tx).
+			PutRecord(ctx /* TODO should i use tx.Statement.Context */, object, object.SpaceOwner(), habitat_syntax.SpaceRelationCollection, "" /* rkey should be generated */, record)
```
sashankg: rm TODOs

internal/perms/store.go
```diff
@@ -118,92 +99,191 @@ func (s *store) AddUserRelation(
...
-	return s.fga.WriteRaw(ctx, &openfgav1.WriteRequest{
-		Deletes: &openfgav1.WriteRequestDeletes{
-			TupleKeys: []*openfgav1.TupleKeyWithoutCondition{
-				tuple.TupleKeyToTupleKeyWithoutCondition(tuple.NewTupleKey(
-					fgastore.SpaceObjectKey(space),
-					fgaRelationFromRole[role],
-					fgastore.MemberUserString(did),
-				)),
+	return s.db.Transaction(func(tx *gorm.DB) error {
+		collection := syntax.NSID(habitat_syntax.UserRelationCollection)
```
sashankg, line 240: we should make the UserRelationCollection const an NSID and downcast it to string when needed

internal/perms/store.go
sashankg: this still needs to be fixed

internal/perms/store.go
```diff
@@ -118,92 +99,191 @@ func (s *store) AddUserRelation(
...
+		records, err := s.spaces.WithTx(tx).ListRecords(ctx, space, space.SpaceOwner(), &collection)
+		if err != nil {
+			return fmt.Errorf("err listing relationship records: %w", err)
+		}
+
+		for _, record := range records {
```
sashankg: discussed offline: if we make the record keys a hash of the record value, then we can look them up more easily here

sashankg (commented):
overall refactor is looking so much cleaner. will approve once the org issue is fixed. if test are passing despite that bug let's add a test

internal/perms/store.go
arushibandi: idk i didn't want this to be doing actual writes

internal/perms/store.go
arushibandi: how? the generated type is a struct but this needs to be a map[string]any ? unless there's a library to do that?

internal/perms/store.go
arushibandi: made this change

sashankg (approved):

PR #837, opened 2026-08-18
[simplespace]: separate out from rest of spaces

arushibandi:
Goal is that `space` package does not depend on `fga`
`simplespace` depends on `space` and `fga`

Soon, `perms` package can depend on `space` and `simplespace` can depend on `perms` instead of `fga` directly.


internal/simplespace/store.go
sashankg, line 1: nit: rename to store.go / manager.go. I personally prefer store

internal/simplespace/simplespace.go
```diff
@@ -0,0 +1,263 @@
...
+
+type space struct {
+	Owner     syntax.DID              `gorm:"primaryKey"`
+	Type      syntax.NSID             `gorm:"primaryKey"`
+	Skey      habitat_syntax.SpaceKey `gorm:"primaryKey"`
+	CreatedAt time.Time
+}
+
+// Methods are analagous to those defined here: https://github.com/bluesky-social/proposals/tree/main/0016-permissioned-data#required-pds-space-management-simplespace
+type Manager interface {
```
sashankg: I've realized this is an anti pattern. The constructor should return the struct pointer directly. callers should be the ones that declare the interface they need.

internal/simplespace/simplespace.go
```diff
@@ -0,0 +1,263 @@
...
+	"github.com/bluesky-social/indigo/atproto/syntax"
+	"github.com/habitat-network/habitat/internal/fgastore"
+	"github.com/habitat-network/habitat/internal/spaces"
+	habitat_syntax "github.com/habitat-network/habitat/internal/syntax"
+	openfgav1 "github.com/openfga/api/proto/openfga/v1"
+	"github.com/openfga/openfga/pkg/tuple"
+	"gorm.io/gorm"
+)
+
+type space struct {
```
sashankg: this is probably fine for now. we might introduce our own endpoint for creating spaces besides `simplespace.createSpace`. `relationship.createSpace` would take params that would let you configure more things than simplespace does. if we do that then we would need a shared table between the relationship store and the simplespace store. I think the simplespace store would just call the relationship store's createSpace since our version of spaces should be a super set of functionality than simplespace.

internal/simplespace/simplespace.go
```diff
@@ -0,0 +1,263 @@
...
+	Skey      habitat_syntax.SpaceKey `gorm:"primaryKey"`
+	CreatedAt time.Time
+}
+
+// Methods are analagous to those defined here: https://github.com/bluesky-social/proposals/tree/main/0016-permissioned-data#required-pds-space-management-simplespace
+type Manager interface {
+	// Space operations
+	CreateSpace(
+		ctx context.Context,
+		org syntax.DID,
```
sashankg: let's use "authority"/"owner" terminology here to be more clear

internal/simplespace/manager.go
```diff
@@ -0,0 +1,263 @@
...
+
+var _ Manager = &manager{}
+
+// WithTx implements [Manager], returning a manager whose DB operations run on tx.
+func (m *manager) WithTx(tx *gorm.DB) Manager {
+	return &manager{
+		db:     tx,
+		spaces: m.spaces,
+		fga:    m.fga,
+		clock:  syntax.NewTIDClock(0 /* TODO: should this be a random number */),
```
sashankg: this should be `m.clock`. We want to maintain atomicity across the entire store even in transactions.

internal/simplespace/simplespace.go
```diff
@@ -0,0 +1,263 @@
...
+var (
+	ErrCannotRemoveOrg    = errors.New("cannot remove the org from the space")
+	ErrSpaceNotFound      = errors.New("space not found")
+	ErrSpaceAlreadyExists = errors.New("space already exists")
+)
+
+// CreateSpace implements [Manager].
+func (m *manager) CreateSpace(ctx context.Context, org syntax.DID, creator syntax.DID, spaceType syntax.NSID, skey habitat_syntax.SpaceKey) (habitat_syntax.SpaceURI, error) {
+	if skey == "" {
+		// TODO: should this / does this need to be a TID?
```
sashankg: no the space key can be any value that satisfies the [record key requirements](https://atproto.com/specs/record-key). the default value is just the current TID

internal/simplespace/manager.go
```diff
@@ -0,0 +1,263 @@
...
+	err := m.db.Transaction(func(tx *gorm.DB) error {
+		if err := tx.Create(&space{
+			Owner: org,
+			Type:  spaceType,
+			Skey:  skey,
+		}).Error; err != nil {
+			if errors.Is(err, gorm.ErrDuplicatedKey) {
+				return ErrSpaceAlreadyExists
+			}
+			return err
```
sashankg: wrap the errors with fmt.Errorf to provide more context

internal/simplespace/store.go
```diff
@@ -0,0 +1,263 @@
...
+			Type:  spaceType,
+			Skey:  skey,
+		}).Error; err != nil {
+			if errors.Is(err, gorm.ErrDuplicatedKey) {
+				return ErrSpaceAlreadyExists
+			}
+			return err
+		}
+
+		return m.fga.Write(
```
sashankg: same wrap the error here and everywhere

internal/simplespace/simplespace.go
```diff
@@ -0,0 +1,263 @@
...
+	var sp space
+	err := m.db.WithContext(ctx).
+		Where("owner = ? AND skey = ?", uri.SpaceOwner(), uri.Skey()).
+		First(&sp).Error
+	if errors.Is(err, gorm.ErrRecordNotFound) {
+		return ErrSpaceNotFound
+	} else if err != nil {
+		return err
+	}
+	if did == uri.SpaceOwner() {
```
sashankg: this should probably be an error

internal/simplespace/simplespace.go
```diff
@@ -0,0 +1,263 @@
...
+	) (habitat_syntax.SpaceURI, error)
+	DeleteSpace(ctx context.Context, uri habitat_syntax.SpaceURI) error
+
+	// Member operations
+	AddMember(ctx context.Context, space habitat_syntax.SpaceURI, did syntax.DID) error
+	RemoveMember(ctx context.Context, space habitat_syntax.SpaceURI, did syntax.DID) error
+	ListMembers(ctx context.Context, callerOrg syntax.DID /* todo: remove this when org membership uses spaces*/, space habitat_syntax.SpaceURI) ([]syntax.DID, error)
+}
+
+type manager struct {
```
sashankg: missing a constructor

sashankg (approved):

internal/simplespace/simplespace.go
arushibandi: can you expand on where you saw this was an anti pattern? i think it depends on each case; we have been declaring too many interfaces probably, but i think in general exporting an interface is okay.

internal/simplespace/store.go
sashankg: this isn't resolved

internal/simplespace/store.go
```diff
@@ -0,0 +1,244 @@
...
+	return uri, nil
+}
+
+// DeleteSpace implements [Store].
+func (m *Store) DeleteSpace(ctx context.Context, uri habitat_syntax.SpaceURI) error {
+	// read the stored FGA tuples for this space before deleting anything,
+	// so we know exactly what tuples to delete
+	tuples, err := m.fga.Read(ctx, fgastore.Tuple{Object: fgastore.SpaceObjectKey(uri)})
+	if err != nil {
+		return err
```
sashankg: wrap

internal/simplespace/store.go
```diff
@@ -0,0 +1,244 @@
...
+
+		// delete all stored FGA tuples for this space
+		var deletes []*openfgav1.TupleKeyWithoutCondition
+		for _, t := range tuples {
+			deletes = append(deletes, tuple.TupleKeyToTupleKeyWithoutCondition(
+				tuple.NewTupleKey(t.Object, t.Relation, t.User),
+			))
+		}
+		if len(deletes) > 0 {
+			return m.fga.WriteRaw(ctx, &openfgav1.WriteRequest{
```
sashankg: wrap

internal/simplespace/store.go
```diff
@@ -0,0 +1,244 @@
...
+				},
+			})
+		}
+		return nil
+	})
+	if err != nil {
+		return fmt.Errorf("err deleting space: %w", err)
+	}
+	// Best-effort: tell registered syncers the space is gone.
+	m.notifier.NotifySpaceDeleted(ctx, uri)
```
sashankg: i feel like this call should go in the space store.

sashankg (approved):

PR #844, opened 2026-08-19
[testing]: flake fix

arushibandi:


internal/fgastore/fgastore_test.go
```diff
@@ -15,9 +16,29 @@ import (
 
 func newTestSQLite(t *testing.T) *FGA {
```
sashankg: let's just use the memory fga in tests and remove this entirely

PR #852, opened 2026-08-20
[relationship server]: user spaces + perms store directly

arushibandi:


internal/relationship/server.go
```diff
@@ -287,38 +332,127 @@ func (s *Server) ListObjects(w http.ResponseWriter, r *http.Request) {
...
+		authn.WithSpace(space, fgastore.RelationSpaceReader),
+	).Validate(w, r); !ok {
+		return
+	}
+	if params.SubjectType != "" && params.SubjectType != "user" && params.SubjectType != "space" {
+		httpx.WriteInvalidRequest(ctx, w, "invalid subjectType", nil)
+		return
+	}
+	if params.Object != "" && params.Object != space.String() {
+		// The object of a relation record is always the space it's written
```
sashankg: we should just remove the object param then, right?

sashankg (approved):

api-docs/static/api.json
```diff
@@ -3979,21 +3979,147 @@
           {
             "name": "subject",
             "in": "query",
-            "description": "The subject to check: a user DID, or a space URI when checking a space-role userset. When a space URI, subjectRole is required.",
+            "description": "URI of the subject space (or group-space) whose role-holders form the userset to check.",
```
arushibandi, line 3982: i think calling it `group-space` is confusing. you can just call it subject space

internal/perms/store.go
```diff
@@ -54,6 +55,15 @@ type Store interface {
 		subjectRole habitat_syntax.SpaceRole,
 		objectSpace habitat_syntax.SpaceURI,
 	) error
+
+	// DeleteRelation removes the relation record at uri (either a userRelation
+	// or a spaceRelation record) from both the governing space and FGA.
+	//
+	// TODO: not yet implemented — resolve uri to the record it points at,
```
arushibandi: remove TODO

internal/perms/store.go
```diff
@@ -431,18 +451,18 @@ func (s *store) ListUserSubjects(
 }
 
 // ListObjects implements [Store]. It returns the spaces did directly holds
-// at least habitat_syntax.SpaceRoleReader on (reader is implied by every other role, so
-// this covers manager/writer/owner too). Owner/org-member implications
-// aren't expanded here, since that would require checking every space in
-// every org did could be a member of.
+// role on (owner/manager/writer imply reader, so a lower role requested
+// returns spaces held via a higher one too). xw
```
arushibandi: i think this is backwards. a higher role requested returns spaces held via a lower one too. als rm xw

internal/relationship/server_test.go
```diff
@@ -14,27 +14,68 @@ import (
...
+	db_testutil "github.com/habitat-network/habitat/internal/db/testutil"
+	"github.com/habitat-network/habitat/internal/fgastore"
 	"github.com/habitat-network/habitat/internal/perms"
 	"github.com/habitat-network/habitat/internal/spaces"
+	spaces_testutil "github.com/habitat-network/habitat/internal/spaces/testutil"
+	habitat_syntax "github.com/habitat-network/habitat/internal/syntax"
 )
 
-func newTestServer(t *testing.T, caller syntax.DID) (*Server, *Store, spaces.Store) {
+var (
```
arushibandi, line 26: would be nice to have a test for simplespace.createSpace , .deleteSpace in conjunction with these APIs, since that is how it is expected to be used

internal/relationship/server_test.go
```diff
@@ -77,93 +130,219 @@ func TestServer_WriteTuple_InvalidSubject(t *testing.T) {
...
 	)
 	w := httptest.NewRecorder()
-	s.DeleteTuple(w, req)
+	s.DeleteRelation(w, req)
 	require.Equal(t, http.StatusBadRequest, w.Code)
 }
 
-func TestServer_ListTuples(t *testing.T) {
-	s, rel, sp := newTestServer(t, org)
+func TestServer_ListRelations(t *testing.T) {
```
arushibandi, line 212: i think we should be testing permissions in this test file as well -- e.g unit test that readers+ above can do all the read queries, only managers + above can set relations for writer + reader or delete relations, only owners can set relations for managers.

internal/relationship/server.go
```diff
@@ -48,44 +55,77 @@ func (s *Server) WriteUserRelation(w http.ResponseWriter, r *http.Request) {
...
+	if !ok {
+		return
+	}
+	if _, ok = s.validator.Request(
+		authn.WithMethods(authn.ValidatorMethodOAuth, authn.ValidatorMethodServiceAuth),
+		authn.WithSpace(space, fgastore.RelationSpaceMemberManager),
+	).Validate(w, r); !ok {
+		return
+	}
+	subjectRole, err := parseSpaceRole(input.SubjectRole)
```
arushibandi, line 121: i think there is some missing authz here. managers can't downgrade/upgrade owners.

internal/relationship/server.go
```diff
@@ -251,25 +299,16 @@ func (s *Server) ListObjects(w http.ResponseWriter, r *http.Request) {
...
-		credInfo.Org.DID(),
-		did,
-		Role(params.Relation),
-		filterType,
-	)
-	if errors.Is(err, ErrInvalidTuple) {
-		httpx.WriteInvalidRequest(ctx, w, "invalid tuple", err)
-		return
-	} else if err != nil {
+	spaceURIs, err := s.perms.ListObjects(ctx, did, role, filterType)
```
arushibandi, line 331: this should be using the did from credInfo no? otherwise it lets anyone read anyone else's objects.

can we have a test for this?

internal/relationship/server.go
```diff
@@ -285,38 +324,126 @@ func (s *Server) ListObjects(w http.ResponseWriter, r *http.Request) {
...
+		return
+	}
+	if params.SubjectType != "" && params.SubjectType != "user" && params.SubjectType != "space" {
+		httpx.WriteInvalidRequest(ctx, w, "invalid subjectType", nil)
+		return
+	}
+
+	views := make([]any, 0)
+
+	if params.SubjectType != "space" {
```
arushibandi, line 380: slighly more confusing reading this than == "user" || == "" but ok

internal/relationship/server.go
sashankg: we filter this list to the ones the caller can see below. we do have a test for this

arushibandi: oh i thought we were gonna change the semantics to make it simpler for now. wdyt.

internal/relationship/server.go
```diff
@@ -59,17 +59,31 @@ func (s *Server) WriteUserRelation(w http.ResponseWriter, r *http.Request) {
...
 	}
+	isSubjectCurrentlyOwner, err := s.perms.CheckUserHasSpaceRole(
+		ctx,
+		subject,
+		space,
+		habitat_syntax.SpaceRoleOwner,
+	)
+	if err != nil {
+		httpx.WriteServerError(ctx, w, fmt.Errorf("check subject is owner: %w", err))
+		return
```
arushibandi, line 82: would be nice to move all of this into authorizeCanWrite, since its duplicated code

just have authorize write take in credInfo, space, settingRole

sashankg: no it's not duplicated. one is `CheckUserHasSpaceRole` and the other is `CheckSpaceRelationHasSpaceRole`

arushibandi: oh. still i feel like it maybe could be cleaner if you do the check for what write they are tryingt to do first, then branch on what you pass into `Validate` after

arushibandi: but we can leave this as a follow up

PR #925, opened 2026-08-28
[chalk]: support org handles

arushibandi:


sashankg:
fixed in https://github.com/habitat-network/habitat/pull/932

PR #930, opened 2026-08-28
[ui]: trim whitespace in login

arushibandi:


sashankg (commented):
let's just trim it at the auth server

sashankg:
Addressed in c129692d: moved handle whitespace trimming out of the frontend forms and into the auth-server boundaries — pear's OAuth `resolveLoginHint` (covers login_hint/handle/disambiguation), `HandlePasswordLogin` (loginMember), and sap's `/session/add` handler. All four UI trim calls were removed.

